PR #9, opened 2015-03-06
Increased default stack size

alexarchambault:
Fixes https://github.com/lihaoyi/scala-parser/issues/8


lihaoyi:
Oh wow, I didn't know you could do this

PR #12, opened 2015-05-10
Readme: @version appearing as literal

japgolly:
> To begin using FastParse, add
> 
> "com.lihaoyi" %% "fastparse" % "0.1.3"
> To your SBT configuration. To use with Scala.js, you'll need
> 
> "com.lihaoyi" %%% "fastparse" % "@fastparse.Constants.version"

`fastparse.Constants.version` not being replaced by value.

### DIFF
--- a/readme/Readme.scalatex
+++ b/readme/Readme.scalatex
@@ -41,9 +41,9 @@
 
         @p
             To your SBT configuration. To use with Scala.js, you'll need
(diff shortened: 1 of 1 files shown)


lihaoyi:
I think I already fixed this in the last 10 minutes, lol. I should go get some sleep

PR #14, opened 2015-05-17
LiteralIgnoreCase - match a string ignoring character case

ttsymlov:
Matches a string ignoring character case.
Example: 
```scala
import fastparse._

val parser = P(IgnoreCase("keyword ").rep)

val Result.Success(value, successIndex) = parser.parse("keyword Keyword keyWord ")

assert(value == (), successIndex == 24)
```

### DIFF
--- a/fastparse/shared/src/main/scala/fastparse/package.scala
+++ b/fastparse/shared/src/main/scala/fastparse/package.scala
@@ -13,6 +13,7 @@ package object fastparse {
   val End = parsers.Terminals.End
   val Index = parsers.Terminals.Index
   val AnyChar = parsers.Terminals.AnyChar
+  val IgnoreCase = parsers.Terminals.LiteralIgnoreCase
 
   val CharPred = Intrinsics.CharPred
   val CharIn = Intrinsics.CharIn
--- a/fastparse/shared/src/main/scala/fastparse/parsers/Terminals.scala
+++ b/fastparse/shared/src/main/scala/fastparse/parsers/Terminals.scala
@@ -69,6 +69,22 @@ object Terminals {
     }
     rec(0)
   }
+
+  def startsWithIgnoreCase(src: String, prefix: String, offset: Int) = {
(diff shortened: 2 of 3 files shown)


ttsymlov:
Hi!
I don't know your plans regarding `Terminals.startsWith`, but we can get rid of `LiteralIgnoreCase` & `startsWithIgnoreCase` adding `ignoreCase: Boolean` parameter to `Terminals.startsWith` and `Terminals.Literal`. In that case, implementation might look like this:
```scala
  def startsWith(src: String, prefix: String, offset: Int, ignoreCase: Boolean = false) = {
    val max = prefix.length
    @tailrec def rec(i: Int): Boolean = {
      if (i >= prefix.length) true
      else if (i + offset >= src.length) false
      else {
        val c1: Char = src.charAt(i + offset)
        val c2: Char = prefix.charAt(i)
        if (c1 != c2 && ignoreCase && c1.toLower != c2.toLower) false
        else rec(i + 1)
      }
    }
    rec(0)
  }

case class Literal(s: String, ignoreCase: Boolean = false) extends Parser[Unit]{
    def parseRec(cfg: ParseCtx, index: Int) = {

      if (startsWith(cfg.input, s, index, ignoreCase)) success(cfg.success, (), index + s.length, false)
      else fail(cfg.failure, index)
    }
    override def toString = literalize(s).toString
  }

val IgnoreCase = (s: String) => parsers.Terminals.LiteralI(s, true)

```

That'll remove copy-paste-look-alike `LiteralIgnoreCase` and `startsWithIgnoreCase` at the cost of cheap `&& ignoreCase` check. On the other hand, `Terminals.startsWith` looks like very hot peace of code. What do you think?

lihaoyi:
This looks great. You're right, the internals are gonna be really hot, so I don't mind some duplication. I'm probably even going to manually specialist the `Either` and `Repeat` parsers, introducing a bit more duplication to let them run faster

PR #15, opened 2015-05-23
fix fastarse.wspStr to actually do return CharLiteral for a single ch…

ttsymlov:
Fix `fastarse.wspStr` to actually do return `CharLiteral` for a single char string


lihaoyi:
lol nice catch. Hopefully this shows up in the benchmarks...

PR #22, opened 2015-06-01
Restrained utest dependency to test scope

alexarchambault:
The utest dependency doesn't get pulled when out of tests, this way.


lihaoyi:
Yay thanks

PR #24, opened 2015-06-02
add filter to API

russwyte:
I implemented filter and a simple test.

I am not certain I handled the predicate failure case as well as I could have. I would appreciate some feedback there.

Thanks!

### DIFF
--- a/fastparse/shared/src/main/scala/fastparse/parsers/Transformers.scala
+++ b/fastparse/shared/src/main/scala/fastparse/parsers/Transformers.scala
@@ -32,16 +32,17 @@ object Transformers {
     }
   }
 
-  case class Filtered[T](p1: fastparse.Parser[T],
+  case class Filtered[T](p: fastparse.Parser[T],
                             predicate: T => Boolean)
     extends fastparse.Parser[T] {
     override def parseRec(cfg: ParseCtx, index: Int): Result[T] = {
-      p1.parseRec(cfg, index) match{
+      p.parseRec(cfg, index) match{
         case f: Result.Failure.Mutable => failMore(f, index, cfg.trace, false)
-        case s: Result.Success.Mutable[T] => {
+        case s: Result.Success.Mutable[T] =>
           if (predicate(s.value)) s else fail(cfg.failure,index,false)
-        }
       }
     }
+
+    override def toString: String = s"$p.filter($predicate)"
   }
 }
--- a/fastparse/shared/src/main/scala/fastparse/parsers/Transformers.scala
+++ b/fastparse/shared/src/main/scala/fastparse/parsers/Transformers.scala
@@ -39,7 +39,7 @@ object Transformers {
       p1.parseRec(cfg, index) match{
         case f: Result.Failure.Mutable => failMore(f, index, cfg.trace, false)
         case s: Result.Success.Mutable[T] => {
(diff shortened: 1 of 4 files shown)


fastparse/shared/src/main/scala/fastparse/parsers/Transformers.scala
```diff
@@ -31,4 +31,17 @@ object Transformers {
...
   }
+
+  case class Filtered[T](p1: fastparse.Parser[T],
+                            predicate: T => Boolean)
+    extends fastparse.Parser[T] {
+    override def parseRec(cfg: ParseCtx, index: Int): Result[T] = {
+      p1.parseRec(cfg, index) match{
+        case f: Result.Failure.Mutable => failMore(f, index, cfg.trace, false)
+        case s: Result.Success.Mutable[T] => {
+          if (predicate(s.value)) s else failMore(cfg.failure,index,false)
```
lihaoyi: This should be `fail` rather than `failMore`, because you're starting the failure propagation not continuing it

fastparse/shared/src/test/scala/fastparse/ExampleTests.scala
```diff
@@ -158,6 +158,16 @@ object ExampleTests extends TestSuite{
           failure.trace == """xml:0 / rightTag:7 / "abcde":9 ..."edcba>""""
         )
       }
+      'filter{
+        val digits = P(CharIn('0' to '9').rep(1).!).map(_.toInt)
+        val even = digits.filter(_ % 2 == 0)
+        val Result.Success(12, _) = even.parse("12")
+        val failure = even.parse("123").asInstanceOf[Result.Failure]
```
lihaoyi: Could we add asserts for `Failure#trace` as well as `even.toString` to make sure they're reasonable? I'd expect `even.toString` to equal `digits.filter(<function1>)` in this case since we're passing in an anonymous lambda

lihaoyi:
that's all for now

russwyte:
Thank you for the feedback! I have changed things according to your suggestions.

fastparse/shared/src/test/scala/fastparse/ExampleTests.scala
lihaoyi: Bump, could we change to `toString` to make it look like `digits.filter(<function1>)`? That's how most people are going to use it, so we should make it look the same ^_^

fastparse/shared/src/main/scala/fastparse/parsers/Transformers.scala
```diff
@@ -31,4 +31,17 @@ object Transformers {
...
     }
   }
+
+  case class Filtered[T](p1: fastparse.Parser[T],
+                            predicate: T => Boolean)
+    extends fastparse.Parser[T] {
+    override def parseRec(cfg: ParseCtx, index: Int): Result[T] = {
+      p1.parseRec(cfg, index) match{
+        case f: Result.Failure.Mutable => failMore(f, index, cfg.trace, false)
+        case s: Result.Success.Mutable[T] => {
```
lihaoyi: Unnecessary curlies; can we remove them?

lihaoyi:
Can you also add it to the readme? A separate PR for the gh-pages site would be nice too if possible...

lihaoyi:
Thanks!

russwyte:
It was fun! Thanks the accept.

PR #29, opened 2015-06-16
Failing test case for the behaviour of `end` in `rep(end=...)`

rtyley:
I was expecting both these parsers to match `"ca"` (perhaps my expectations about `end` are incorrect, if so, it would be great to add clarification):

* `CharIn("abc").rep(end=P(&("XYZ")))` matching on `"caXYZaa"`
* `CharIn("abc").rep(end=P(&("cba")))` matching on `"cacbaaa"`

...but currently FastParse fails on the second of the two checks:

```
[info]       		utest.AssertionError: assert({parser; str; parsed} == rhs)
[info]       		parser: fastparse.P[T] = CharIn("abc").rep(end = tests)
[info]       		str: String = cacbaaa
[info]       		parsed: fastparse.core.Result[T] = Failure((CharIn("abc") | tests):7 ..."", false)
[info]       		rhs: fastparse.core.Result[T] = Success(List(c, a), 2)
```

(Incidentally, I did have an unsuccessful go at implementing a corresponding fix, in `Repeater`, but fell over in `NullPointerException`s, of all things)

### DIFF
--- a/fastparse/shared/src/test/scala/fastparse/ParsingTests.scala
+++ b/fastparse/shared/src/test/scala/fastparse/ParsingTests.scala
@@ -42,6 +42,8 @@ object ParsingTests extends TestSuite{
       check("Hello".!.rep, ("HelloHello!", 2), Success.Mutable(Seq(), 2))
       check("Hello".!.rep, ("HelloHello!", 5), Success.Mutable(Seq("Hello"), 10))
       check("Hello".!.rep(1), ("HelloHello!", 0), Success.Mutable(Seq("Hello", "Hello"), 10))
+      check(CharIn("abc").!.rep(end=P(&("XYZ"))), ("caXYZaa", 0), Success.Mutable(Seq("c","a"), 2))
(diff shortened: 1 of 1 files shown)


lihaoyi:
Your first case seems to match just fine

Your second case should not match, at least with the current (under-specified) semantics: it checks the body of the `rep` before checking the `end`, and therefore it will keep running. You can basically treat `end=p` as a `~ p` with better error reporting.

This behavior is kind of awkward, and not very intuitive. The only reason `end` exists is to improve error reporting; it's not really documented anywhere and I hope to get rid of it eventually. For now it's still here.

rtyley:
> The only reason end exists is to improve error reporting; it's not really documented anywhere and I hope to get rid of it eventually.

Fair enough, I wasn't sure if `end` was blessed or not! In the end I used a negative look-ahead to get the effect I wanted (only consuming lines until a new header was encountered):

https://github.com/rtyley/submitgit/blob/93d4f96f/app/lib/model/Patch.scala#L22

lihaoyi:
Cool, sounds like we have a solution then =)

PR #30, opened 2015-06-18
Moved FuncName to top-level

drdozer:
FuncName and its companion was nested inside Util, and the companion referred to reflection types. This caused run-time references to FuncName to cause class-loading of the reflection classes, which shouldn't be needed or even available at run-time.

I've fixed this by lifting FuncName up to the top level so that it no longer can potentially trigger a class-load of Utils as its outer class. I've also moved the impl function that implements the FuncName summoning macro out of the FuncName companion into Utils, again so that all the FuncName bytecode can be loaded without triggering loading of reflection.

### DIFF
--- a/fastparse/js/src/main/scala/fastparse/CharPredicates.scala
+++ b/fastparse/js/src/main/scala/fastparse/CharPredicates.scala
@@ -2,18 +2,18 @@ package fastparse
 
 object CharPredicates{
   // Not available in Scala.js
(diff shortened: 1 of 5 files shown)


drdozer:
I think I've caught the remaining cases now.

lihaoyi:
I manually patched this down in https://github.com/lihaoyi/fastparse/commit/98eced77bbf366423c02c27aa20cc75916617262

PR #39, opened 2015-09-21
Use CharSequence instead of String for more flexibility

### LINKED ISSUE #18: Parse CharSequence rather than String
That would make it more general and let Ammonite parse `Vector[Char]`s without mangling them into strings first

cvogt:
This e.g. allows using a lazy sequence for streamed parsing. Replaced eager .length calls with less eager charAt calls where needed. Added test that asserts laziness.

fixes #18

### DIFF
--- a/fastparse/shared/src/main/scala/fastparse/parsers/Intrinsics.scala
+++ b/fastparse/shared/src/main/scala/fastparse/parsers/Intrinsics.scala
@@ -3,6 +3,7 @@ import acyclic.file
 import fastparse.Utils._
 import fastparse.core.{Precedence, ParseCtx, Result, Parser}
 import fastparse.Utils
+import scala.util.Try
 
 /**
  * High-performance intrinsics for parsing common patterns. All
@@ -16,8 +17,8 @@ object Intrinsics {
     private[this] val uberSet = CharBitSet(chars)
     def parseRec(cfg: ParseCtx, index: Int) = {
       val input = cfg.input
-      if (index >= input.length) fail(cfg.failure, index)
-      else if (uberSet(input(index))) success(cfg.success, (), index + 1, Nil, false)
+      if (Try(input.charAt(index)).isFailure) fail(cfg.failure, index)
+      else if (uberSet(input.charAt(index))) success(cfg.success, (), index + 1, Nil, false)
       else fail(cfg.failure, index)
     }
   }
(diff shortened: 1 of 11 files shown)


fastparse/shared/src/main/scala/fastparse/parsers/Intrinsics.scala
```diff
@@ -16,8 +17,8 @@ object Intrinsics {
     private[this] val uberSet = CharBitSet(chars)
     def parseRec(cfg: ParseCtx, index: Int) = {
       val input = cfg.input
-      if (index >= input.length) fail(cfg.failure, index)
-      else if (uberSet(input(index))) success(cfg.success, (), index + 1, Nil, false)
+      if (Try(input.charAt(index)).isFailure) fail(cfg.failure, index)
+      else if (uberSet(input.charAt(index))) success(cfg.success, (), index + 1, Nil, false)
```
cvogt: @lihaoyi not completely happy with having an intermediate exception being generated here, but .length is too eager. any better ideas?

fastparse/shared/src/main/scala/fastparse/parsers/Terminals.scala
```diff
@@ -31,7 +32,7 @@ object Terminals {
     def parseRec(cfg: ParseCtx, index: Int) = {
       val input = cfg.input
       if (index >= input.length) fail(cfg.failure, index)
-      else success(cfg.success, input(index), index+1, Nil, false)
+      else success(cfg.success, input.charAt(index), index+1, Nil, false)
```
cvogt: Are there any semantic differences between charAt and Scala's apply on Strings?

cvogt:
*bump* :)

lihaoyi:
Sorry, haven't taken a serious look yet!

The `Try(input.charAt(index)).isFailure` pattern probably won't do; that will cut into performance a great deal will all the extra allocations of `Try`s and thunks. I don't know what a better solution would be. Perhaps using our own custom structs instead of `CharSequence`/`Stream[Char]` that provides more convenient/performant access to check if something is available?

cvogt:
@lihaoyi yeah, maybe we need our own CharSequence :-\. Would be good to measure though first what the performace impact of the Try or maybe using try{}catch{} actually is. You have performance tests, right?

lihaoyi:
Yeah just run https://github.com/lihaoyi/fastparse/blob/master/scalaparse/jvm/src/test/scala/scalaparse/PerfTests.scala and see how long it takes

lihaoyi:
Closing due to inactivity

PR #44, opened 2015-10-23
Added example of `NoCut` to documentation.

Ichoran:
This ought to document the use of `NoCut` to enable better composition of independently written parsers.  ("Ought to" because not all tests run as I'm not set up for scala.js, and I am entirely unfamiliar with Scalatex.)


Ichoran:
If the build failure has anything to do with this PR, it is not obvious to me.

lihaoyi:
Looks like a flake, should be fine

PR #49, opened 2015-11-03
Add upper bound to repeat operator (fix #27)

### LINKED ISSUE #27: Add an upper boundary to rep
It's currently possible to specify a minimum number of repetitions for repeated sequences. It'd be great to also have a maximum number of repetitions.

My use case is parsing HTTP grammar, where for example [language tags](http://tools.ietf.org/html/rfc2616#section-3.10) are defined as:

    language-tag  = primary-tag *( "-" subtag )
    primary-tag   = 1*8ALPHA
    subtag        = 1*8ALPHA

I haven't yet found a sane way to express that with fastparse, but maybe I'm missing something?

Alain-Bearez:


Alain-Bearez:
I don't understand what the error log means at https://travis-ci.org/lihaoyi/fastparse/jobs/88978992#L3108

lihaoyi:
Looks like just a flake

fastparse/shared/src/test/scala/fastparse/ExampleTests.scala
```diff
@@ -43,8 +43,14 @@ object ExampleTests extends TestSuite{
         val abc = P( "a".rep(sep="b") ~ "c").log("A")
```
lihaoyi: No need the `log` in the test suite

fastparse/shared/src/test/scala/fastparse/ExampleTests.scala
```diff
@@ -43,8 +43,14 @@ object ExampleTests extends TestSuite{
...
         val Result.Success(_, 8) = abc.parse("abababac")
-
         val Result.Failure(parser, 3) = abc.parse("abaabac")
+
+        val ab4c = P ( "a".rep(min=2, max=4, sep="b") ~ "c" )
+
+        val Result.Failure(_, 1) = ab4c.parse("ac")
+        val Result.Success(_, 4) = ab4c.parse("abac")
+        val Result.Success(_, 8) = ab4c.parse("abababac")
+        val Result.Failure(_, 7) = ab4c.parse("ababababac")
```
lihaoyi: Can we have one test case for

```scala
val ab4 = P ( "a".rep(min=2, max=4, sep="b") )
val Result.Success("abababa", 8) = ab4.parse("ababababababa")
```

lihaoyi:
That is all. Thanks for the PR =)

fastparse/shared/src/test/scala/fastparse/ExampleTests.scala
Alain-Bearez: I did not touch this line. Yet. But I'll remove the log in my next commit.

lihaoyi: Ah, ok it's my fault then. Just leave it in that case, I'll clean it up

lihaoyi:
Oh I didn't mean to get rid of the other cases! Just to add that one on top of them.

Alain-Bearez:
Sorry! I misunderstood. I will push the ab4c example back.

build.sbt
```diff
@@ -18,14 +18,15 @@ val shared = Seq(
...
+    "com.lihaoyi" %%% "utest" % "0.3.1" % "test"
   ),
   scalaJSStage in Global := FullOptStage,
   organization := "com.lihaoyi",
   version := _root_.fastparse.Constants.version,
-  scalaVersion := "2.11.6",
-  libraryDependencies += "com.lihaoyi" %% "acyclic" % "0.1.2" % "provided",
-  addCompilerPlugin("com.lihaoyi" %% "acyclic" % "0.1.2"),
+  scalaVersion := "2.11.7",
+  ivyScala := ivyScala.value map { _.copy(overrideScalaVersion = true) },
```
lihaoyi: What does this do? It's causing the merge conflict and I have never seen this before

lihaoyi:
Can you rebase this on top of master?

build.sbt
Alain-Bearez: it was a suggestion of sbt when it complained for several different versions of Scala being used by the modules. I just did a clean build without this line and the warning does not appear anymore.

lihaoyi:
Looks like a real breakage

lihaoyi:
Skipping it works for me :+1:

fastparse/shared/src/main/scala/fastparse/parsers/Combinators.scala
```diff
@@ -341,43 +342,50 @@ object Combinators {
...
-      def passIfMin(cut: Boolean,
+      def passInRange(cut: Boolean,
                     lastFailure: Mutable.Failure,
                     finalIndex: Int,
                     acc: R,
                     count: Int) = {
-        if (count >= min) success(cfg.success, acc, finalIndex, lastFailure.traceParsers, cut)
-        else  failMore(lastFailure, index, cfg.logDepth, cut = cut)
-
+        if ((min <= count) && (count <= max)) {
```
lihaoyi: I'm pretty sure the `(count <= max)` is unnecessary because the tail-recursion stops when `count == max`, so both call sites above will never see this condition become true

lihaoyi:
That should *really* be all. Sorry for so much back and forth!

fastparse/shared/src/main/scala/fastparse/parsers/Combinators.scala
Alain-Bearez: Sure! I forgot to remove this check (which did not stop the recursion) after I did it correctly.

lihaoyi:
Merged, Thanks!!!

Issue #46, opened 2015-11-01
Add helpers to get line/col of error message


lihaoyi:
`index` is enough for machines but hard to read for humans. It's trivial to get the line/col via

```scala
val lines = f.input.take(f.index).lines.toVector
val line = lines.length 
val col = lines.last.length
```

But it's probably worth chucking this (or some more-optimized version of it) into the `Failure` object for everybody's convenience

Alain-Bearez:
Would you want to also report the parser position as ```line:column``` in ```Failure.formatParser```? Most calls to this function are not enclosed in the scope of a ```Failure```.

It might be the case to push some additional information into the ```ParseCtx```, like the number of already read lines and the global input index at start of current line.

lihaoyi:
reporting `line:column` in `formatParser` seems reasonable. I'm not sure if we still want to show the `index` somewhere; it seems like it might be useful but maybe it isn't

You shouldn't need too much additional info; IIRC most places where we call `formatParser` already have access to the `input: String`, and that + `index` is enough to re-derive the line and column

lihaoyi:
This has landed in #52. Thanks @Alain-Bearez!

PR #52, opened 2015-11-05
Report line:column instead of index, on failure (#46)

Alain-Bearez:


fastparse/shared/src/main/scala/fastparse/core/Parsing.scala
```diff
@@ -78,13 +78,16 @@ object Result{
...
+      Nil, input, index, Failure.formatParser(lastParser, input, index)
     )
     override def toString = s"Failure($msg)"
   }
   object Failure {
-    def formatParser(p: Precedence, i: Int) = {
-      Precedence.opWrap(p, Precedence.`:`) + ":" + i
+    def formatParser(p: Precedence, input: String, index: Int) = {
+      val lines = input.take(1 + index).lines.toVector
+      val line = lines.length
```
lihaoyi: While you're at this, can you add `line` and `col` onto the `Failure` class as a lazy val?

fastparse/shared/src/main/scala/fastparse/core/Parsing.scala
```diff
@@ -152,8 +159,12 @@ object Result{
      * the default error message isn't to your liking.
      */
     lazy val trace = {
-      Failure.formatStackTrace(stack, input, index, Failure.formatParser(expected0, index))
+      Failure.formatStackTrace(stack, input, index, Failure.formatParser(expected0, input, index))
     }
+
+    private lazy val lines = input.take(1 + index).lines.toVector
```
lihaoyi: No need to have it on `TracedFailure`; you already have it on `Failure`, and if you have a `TracedFailure` you have a `Failure`

lihaoyi:
can you rebase this? I'm not sure what workflow you're using that causes all these merge errors, but it sure is making a lot of them =P

Alain-Bearez:
I started four distinct branches in order to tackle four distinct issues. Each time you merge one of them, the changes I had to replicate through all of them, such as excluding the SocketHandler from lila, generate a conflict with the same change based on what is now a different parent branch.

I don't mind doing a last rebase once you reviewed the changes up to your expectations.

lihaoyi:
This looks good to me! Thanks =D

Issue #42, opened 2015-10-19
Cuts are visually indistinct and typo-prone


Ichoran:
`~! "foo"` and `~!" foo"` and `~ !"foo"` all parse, and all do different things, despite being nearly indistiguishable visually.

Either negation or the cut operator should change.  Negation could be `Not`, or cut could be `~>` or somesuch.  (`~>` because it visually suggests it's one way, and it parses with the same precedence as `~!`.)

lihaoyi:
I understand the problem with `~!`. 

OTOH I'm also reluctant to make it `~>`; that re-uses an operator already defined in scala-parser-combinators and atto, and pretty widely used, to do something totally different. That's not great for people coming from the other two libraries, which I'd guess are a non-trivial number.

Basically anything starting with `~` will have the correct precedence; maybe something like `~/`? I don't actually know if it's worth changing, but maybe that's my thousands-of-lines-of-scala-parser-combinator-experience speaking...

Ichoran:
Well, almost anything would be better than `~!`, since your writeup made it seem like a really good idea to use cuts, and I used them, and then was immediately bitten by the difficulty of reading and not typoing the syntax.  `~/` would be okay.  `~#` is faster to type and is more visually distinct.  `~~~` is also pretty fast to type and visually clear, though it's not so obvious it might mean no backtracking.

Alain-Bearez:
What about ```~!~``` which would keep the idea of the classical sequence operator, only cut in the middle. This way any whitespace on either side of the central ```!``` would lead to a compile error.

Alain-Bearez:
I implemented this alternate cut sequence operator in pull request #53, for you to check.

lihaoyi:
Closed by #57

PR #53, opened 2015-11-06
Change cut sequence operator to ~!~ to avoid human misreadings (issue #42)

Alain-Bearez:


lihaoyi:
While we're bike-shedding, can we use `~/` instead of `~!~`? I think it's better in two ways:

- Saves one character on a very common operation
- Does not use the same character as `!` which helps avoid the confusion this issue was meant to solve

Ichoran:
@lihaoyi - I'm fine with ~/ as it solves the problem, but it occurs to me that ~# is faster to type and more visually distinct. Is there any point in considering that one?  Maybe I shouldn't be writing parsers such that my typing speed on cuts is the limiting factor...but actually with the last few it has been typing-limited overall (not just on cuts).

@Alain-Bearez - Thanks for taking initiative on this!

lihaoyi:
I don't like `~#` because I don't like using `#` in operators at all. It's already a keyword in Scala, and a keyword (line-comment) in many other languages =P

I can't see why you'd be typing limited... How big are your parsers? The Scala parser is <1000LOC and I don't see what your parsing that is more complex than the Scala syntax. What kinds of grammars are we talking about here? Are you missing out on potential de-duplication via sub-parsers and functions? Even C&P would avoid the typing-too-much problem...

lihaoyi:
Also you must be waaaaaaay better at writing grammars than I am if you're not spending an inordinate amount of time debugging these things. I'm looking at somewhere around 95% of the time debugging after spending 5% writing the first grammar out fully

Ichoran:
Okay, fair enough.  ~/ is fine with me.  (Except for you not liking `#`, I think the same arguments apply for `/` at least if it makes you think of `//`, but I don't really care as long as it's not `~!`.)

{ "aside": "And it's not because I'm better at writing parsers--it's because my parsers are all tiny.  I think a subdialect of JSON is the longest, and it's under 100 LOC.  In some of the cases I could probably even use a single regex instead, except the regex is only barely comprehensible and the parser is super-clear.  For instance, _C. elegans_ strains are two or three capital letters followed by an integer, but that column of text that I'm parsing sometimes has an extra letter or something (yay, manual unvalidated text entry), and/or a space and description.  But that's one of ten or so columns.  And that's just for that _particular_ data file.  And sometimes people entered the wrong thing in a particular column, so you want to try parsing the likely wrong things also, etc..  It is, sadly, not interestingly complex enough (like a language) to need much thought or debugging." }

lihaoyi:
Looks like a real failure

Alain-Bearez:
For some (still) obscure reason, the *"second half"* of the refactoring is failing. This is when I try to replace the ```~!``` without parameter with ```~/```.

I will give it yet another try, this time by fresh branching out of current master. This branch now has way too many commits for such a simple purpose: replace ```~!``` by ```~/```.

Alain-Bearez:
Since pull request #56 is successfully passing the integration tests, I will now close this PR.

Issue #33, opened 2015-08-06
StringIn takes a very long time with longer strings


tannerezell:
Tested on 0.2.1:
```scala
sealed trait Val
case class Var(value : String) extends Val
val startTime = new Date()
val vars = List("\r\n", "\n", "hl", "SA", "avxavxavxavxavxavx").toSeq 
val variables : Parser[Var] = P ( StringIn(vars:_*).! ).map(Var)
val Result.Success(myVars, _) = variables.parse("SA")
val endTime = new Date()
val dateDiff = new SimpleDateFormat("mm:ss").format(new Date(endTime.getTime - startTime.getTime))
println (s"myVars = $myVars")
println (s"took $dateDiff to parse")
```

results in the following output:
```
myVars = Var(SA)
took 03:11 to parse
```

This only happens on long strings, shorter ones process almost instantly. I've tried various combinations of the long string and it always results in an exceedingly long parsing time.

lihaoyi:
I wonder if it's a consequence of the way StringIn is implemented? A lot of
our bitsets are implemented as Char predicates, including those used in
StringIn, and those initialize a 65000-pointer long lookup array the first
time they're initialized. This may be causing the slowdown, but I can't
imagine it taking 3 minutes...

Could you repeat the benchmark moving the initialization (construction of
the parser object) outside of the timing block? I thought the
initialization would be relatively fast, but maybe it's not, and we should
reconsider this approach.

The lookup array implementation lives in the trie here:

https://github.com/lihaoyi/fastparse/blob/master/utils/shared/src/main/scala/fastparse/Utils.scala#L138-L183

tannerezell:
It looks like when parse is called, rather than when the parser is setup:

```scala
sealed trait Val
case class Var(value : String) extends Val

def time[R](unit : => R) : R = {
    val startTime = new Date()
    val result = unit
    val endTime = new Date()
    val dateDiff = new SimpleDateFormat("mm:ss:SS").format(new Date(endTime.getTime - startTime.getTime))
    println (s"took $dateDiff to parse")
    unit
  }

  
  val vars = List("\r\n", "\n", "hl", "SA", "avxavxavxavxavxavx").toSeq
  val variables : Parser[Var] = time { P ( StringIn(vars:_*).! ).map(Var) }
  val Result.Success(myVars, _) = time { variables.parse("SA") }
  println (s"myVars = $myVars")
```

and the result:

```
took 00:00:01 to parse
took 03:27:331 to parse
[...]

PR #54, opened 2015-11-06
Fast initialization strategy for small strings set in StringIn (for #33)

Alain-Bearez:


lihaoyi:
Seems pretty ad hoc; what's the purpose of this?

My ideal solution would, I think, be implementing something like a 

http://linux.thai.net/~thep/datrie/datrie.html

To have a fast, compact matcher. But that would be more work...

Alain-Bearez:
I will not have time to work seriously on this subject until the end of the year.

Closing this attempt for now...

lihaoyi:
Ok no problem, thanks for the attempt =) And all the other fixes you send in!

Issue #34, opened 2015-08-12
Pattern-matching on Result produces a (spurious?) warning


jducoeur:
When my code tries to match on the result of a parse, such as:
```
    dumpfileP.parse(mySQL) match {
      case Result.Success(stmts, _) => stmts
      case Result.Failure(parser, index) => displayFailure(...)
    }
```
The Scala compiler consistently kicks out a warning:
```
[warn] /home/jducoeur/GitHub/Querki/querki/scalajvm/app/querki/imexport/MySQLImport.scala:143: The outer reference in this type test cannot be checked at run time.
[warn]       case Result.Success(stmts, _) => stmts
```
This warning may be correct -- I'm honestly unsure -- but I don't care about it, and definitely don't want to see it.  My coding standards are "no warnings", so this is getting in the way of using FastParse for production code.

This can be worked around by using isInstanceOf + asInstanceOf, but that's rather boilerplatey.  A more idiomatic-Scala solution would be preferable.

jducoeur:
Note that adding
```
import Result._
```
and removing the explicit `Result.` in the pattern matches has no effect -- the outer reference is still the same, regardless of whether it is explicit or not.

lihaoyi:
What about

```scala
val (Success, Failure) = (Result.Success, Result.Failure)
```

=D

jducoeur:
Interesting suggestion, but when I try:
```
    val (mySuccess, myFailure) = (Result.Success, Result.Failure)
    dumpfileP.parse(mySQL) match {
      case mySuccess(stmts, _) => stmts
      case myFailure(parser, index) => ...
    }
```
I'm still getting the same warning.  The issue must have to do with the Type itself, not how it's accessed.

jducoeur:
What's really weird here is that this is a [known issue](https://issues.scala-lang.org/browse/SI-4440) in Scala -- but only when the inner case class is marked final.  In this case, it's not.  So I don't know why this is happening...

jducoeur:
Effing fascinating.  I'm experimenting with creating a minimal case for this problem, and far as I can tell, the issue isn't the classes at all -- it's the way it's exporting them.  I couldn't repro the problem with simplified code until I introduced `fastparse.all`.  Continuing to experiment...

jducoeur:
Okay, here's a more or less minimalish example that shows the warning:
```
package querki.test {
  
  import fakefastparse.all._
  
  class FastParseTests extends QuerkiTests {
[...]

PR #55, opened 2015-11-07
build tests to reproduce issue #34

Alain-Bearez:
I was trying to write failing tests in order to reproduce the issue...


lihaoyi:
I don't quite follow. Are these tests meant to fail? Or do they just print warnings?

Alain-Bearez:
Sorry! It should not have been made a pull request. It was late and should have stayed away from the computer.

PR #59, opened 2015-11-19
f.c.R.Failure.formatParser throws an exception on an empty input string

solar:
fastparse.core.Result.Failure.formatParser() throws an UnsupportedOperationException on receiving an empty string as an input.

```
import fastparse.all._

val p = P("a")
p.parse("")
p.toString // throws
```


lihaoyi:
Looks great. Do you think you could add a line to the changelog? =)

solar:
In the changelog of 0.3.2?

lihaoyi:
0.3.3, since 0.3.2 is already out

solar:
Ok, fixed.

lihaoyi:
Looks good. Thanks!

PR #61, opened 2015-11-22
Add Opaque combinator

stanch:
As discussed here: https://gitter.im/lihaoyi/fastparse?at=564e1994b0853133165607bc.

### DIFF
--- a/fastparse/shared/src/main/scala/fastparse/ParserApi.scala
+++ b/fastparse/shared/src/main/scala/fastparse/ParserApi.scala
@@ -13,6 +13,13 @@ trait ParserApi[+T] {
    * where a parser was tried and its result, which is useful for debugging
    */
   def log(msg: String = this.toString)(implicit output: Logger): Parser[T]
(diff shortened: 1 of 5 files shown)


lihaoyi:
Can we make an un-specified `msg` (`null`? `""`? Idunno) return the `toString` of the parser? I imagine that'll be useful in a bunch of cases where your `P(...) ` parser already has a reasonable name, and you just want to hide the insides. In general, FastParse's methods are chock-full of helpers of various sorts, so it would fit right in.

Also, could we add a line to the readme, `sect.ref`-ed to the header in the docs? =)

lihaoyi:
And by "readme" of course I mean "changelog"...

stanch:
You can already do this: `parser.opaque()`, which will use `parser.toString`. I was trying to reduce the number of places where there are defaults.

I will update the changelog.

lihaoyi:
Merged. Thanks @stanch!

PR #62, opened 2015-11-23
Failure and Extra as ordinary case classes.

MartinSenne:
Refactored `Failure` to

    case class Failure(lastParser: Parser[_],
                     index: Int,
                     extra: Extra)

and introduced 

    case class Extra(input: String, traceData: (Int, Parser[_]) )(index: Int, lastParser: Parser[_])

as discussed on Gitter. The currying in `Extra` is required as not to expose `index` and `lastParser` again, but we have to keep a copy in order to have all parameters available in `Extra` to construct a `TracedFailure`.

Tests via `modules/test` run through.
Checked Documentation, that requires no urgent modifications.

Documentation should be changed eventually to include that

* input is available via `Extra`
* `.col` and `.line` are available via `Extra`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,9 +6,41 @@ to use Fastparse, you probably will want to check out the documentation:
 
 - [Documentation](https://lihaoyi.github.io/fastparse)
 
+This readme contains some developer docs, if you intend on working on the fastparse repo, not just using it as a library.
+
+Developer Docs
+==============
+
+The core of FastParse lives in the `fastparse/` folder. It is cross-built ScalaJVM/Scala.js codebase, with almost everything shared between the two platforms in the `fastparse/shared/` and minor differences in `fastparse/js/` and `fastparse/jvm/`.
+
+The two subprojects `scalaparse/` and `pythonparse/` are FastParse parsers for those respective languages. These are both usable as standalone libraries, and also serve as extensive test-suites and use-cases for FastParse itself. Each of those projects clones & parses large quantities of code from Github as part of *their* own test suites.
+
+`util/` contains some basic utilities that FastParse uses that aren't specific to parsers: bitsets, escaping, tries, etc.. `readme/` contains the documentation site, which contains several live demos of FastParse parsers compiled to Scala.js. These all live in `demo/`.
+
+Common Commands
+---------------
+
+- `sbt ~fastparseJVM/test` runs the main testsuite. If you're hacking on FastParse, this is often where you want to go
+- You can run the other suites for `fastparseJS`, `scalaparseJVM`, etc. if you wish, but I typically don't and leave that to CI unless I'm actively working on the sub-project
+- You can use `+` to run it under different Scala versions, but again I usually don't bother
+- `+modules/test` is the aggregate test-all command, and `+modules/publishSigned` is publish-all. Other things (`compile`, etc.) can also be run on `modules`
+- `readme/run` builds the documentation site, which can then be found at `readme/target/scalatex/index.html`
+
+Contribution Guidelines
+-----------------------
+
+- **If you're not sure if something is a bug or not, ask on Gitter first =)**
+- **All code PRs should come with**: a meaningful description, inline comments for important things, unit tests, and a green build
+- **Non-trivial changes, including bug fixes, should appear in the changelog**. Feel free to add your name and link to your github profile!
+- **New features should be added to the relevant parts of the documentation**
+- **To a large extent, FastParse is designed so that you can extend it in your own code** without needing to modify the core. If you want to add features, be prepared to argue why it should be built-in and not just part of your own code.
+- **It's entirely possible your changes won't be merged**, or will get ripped out later. This is also the case for my changes, as the Author!
+- **Even a rejected/reverted PR is valuable**! It helps explore the solution space, and know what works and what doesn't. For every line in the repo, at least three lines were tried, committed, and reverted/refactored, and more than 10 were tried without committing.
+- **Feel free to send Proof-Of-Concept PRs** that you don't intend to get merged.
+- **No binary or source compatibility is guaranteed between any releases**. FastParse is still in the 0.x.y phase of development, which means it's still under rapid development and things do change.
+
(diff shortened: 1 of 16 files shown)


lihaoyi:
The current setup is kind of messy. We're hiding data in `Extra` in two different ways; one in a public but kinda-not-meant-to-be-used `tracedData` field two via that currying thing.

AFAICT these are our goals/constraints:

- `Extra` should not publicly leak anything that's already available on `Failure`: there should be only one way for a user to get each "thing" he wants.
- `Failure` should not have too much extra stuff polluting it's API, especially for the common case where `index`+`lastParser` is enough
- `Extra` needs to contain the lazy `TracedFailure` object, and thus needs enough info to compute it, some of which is redundant on `Failure`

IMHO we should make 

```scala
case class Extra(startIndex: Int, startParser: Parser[_], input: String, lastParser: Parser[_]) 
extends (() => TracedData)
```

And then make `Failure`

```scala
case class Failure(lastParser: Parser[_], index: Int, traced: () => TracedData)
```

That will give us some nice properties:

- `Extra` is just a normal case class without all the weird hiding stuff
- `Failure` doesn't leak out `Extra` at all. It's encapsulated behind the lambda `traced`, which can cache it's result.
- Pattern matching is nice; instead of 
```scala
val Failure(_, _, extra) = foo
extra.traced
``` 

we can say 

```scala
val Failure(_, _, traced) = foo
traced()
```

- Case-class equality continues to work, since even though `Extra` is typed as a function it'll still use its own case-class equality. That means identical failures are still equal.

@MartinSenne WDYT?

lihaoyi:
Lastly, can you insert a line into the changelog for 0.3.3? =D

MartinSenne:
1. If we have

    case class Failure(lastParser: Parser[_], index: Int, traced: () => TracedData)

how do we access the data in Extra then? Or do you mean

    case class Failure(lastParser: Parser[_], index: Int, traced: Extra)

As `Extra` derives from `Function0` we can then call `traced()` (function application).

2. What is this `tracedData` in original `Failure` with (Int, Parser[_]) for? Doea it contain the startIndex and startParser?

lihaoyi:
> how do we access the data in Extra then? Or do you mean

Good question. Is there data we want to expose? If the only thing we're ever gonna do with `extra` is called `.traced` on it, `() => TracedData` is sufficient. If there's additional data we want to expose, we could hide it behind an interface so all the other random stuff doesn't like out.

> Second, what is this tracedData in original Failure with (Int, Parser[_]) for? Doea it contain the startIndex and startParser?

Yeah that's what i contains. It's what's used to do the tracing right now, IIRC.

MartinSenne:
Ok, let me clarify (now I'm on my machine and I need not to type in via mobile)

## In reply to requirements and your comments

* What is `TracedData`? We do not have this as a data type.
  * Do you mean `TracedFailure`?
  * Or do you refer to `traceData: (Int, Parser[_]` which represents `startIndex` and `startParser`? But why do we need an extra datatype then?

> `Extra` should not publicly leak anything that's already available on Failure: there should be only one way for a user to get each "thing" he wants.

Exactly, that is the way it is. No way to access the curried information in `Extra` from outside.

> Failure should not have too much extra stuff polluting it's API, especially for the common case where index+lastParser is enough

Totally agree!

> Extra needs to contain the lazy TracedFailure object, and thus needs enough info to compute it, some of which is redundant on Failure

Exactly !

# Proposal

Let us determine the API:

1. `Failure` case class with

  ```
  case class Failure(lastParser: Parser[_], index: Int, traced: Extra)
  ```

  * Easy access to `lastParser` and `index` . All other information (TraceFailure, startIndex, ... ) is available via `Extra`. **Agree?**

2. All additional or supplementary information is put into `Extra`. 

  ```
  case class Extra(startIndex: Int, startParser: Parser[_], input: String) 
  extends (() => TracedFailure)
  ```

  * `startIndex` and `startParser` as in `tracedData`
  * `TracedFailure` which can be accessed via function application. (formerly accessible via `.traced`
  * `input` as the input to parse
  * **Issues** `lastParser` and `index` are available in `Failure` only. But we need to somehow have them available in `Extra` to **construct** the `TracedFailure` and without making them publicly accessible. (That was the reason I made them curried in my approach, where the final `Extra` is a regular case class BTW. And I see no better approach to circumvent this.)
**Agree?**

MartinSenne:
So, we now have 
```
case class Failure(lastParser: Parser[_],
                     index: Int,
                     input: String,
                     traced: Traced)
```
which still has 4 parameters, but this is acceptable, I think. (All other things I tried were somewhat feeling smelly). Especially, since pulling out `input` requires some additional place for the storage or it means to alter `Traced`. So I decided, to keep it where it is.

`Traced` has become a (interface) trait (with an apply method) as discussed

```
trait Traced extends Function0[TracedFailure] {

    /** Call to obtain the [[TracedFailure]]. This is a costly operation (3-4 times the time it takes to execute [[Parser.parse()]]. */
    def apply() : TracedFailure
  }
```

A also refactoring the `line` and `column` computation, especially as the some code appeared twice in Parsing.scala.

Hope your are happy ;)

One minor things:

* If we really want a simple match for the two parameters index and lastParser of Failure we can do
```
object FailureInfo {
  def unapply(f: Failure) : Option[(Int, String) = Some( (f.index, f.lastParser) )
}

// allows us to to
case FailureInfo(idx, p) => .....
```

lihaoyi:
Tentatively merged as https://github.com/lihaoyi/fastparse/commit/2654485dda72dfeed4bbfdae65ad847bc3f9b3e4, with some changes that kinda hybridize all the different ideas that have come up so far. I think it's turned out pretty nice! Take a look at that diff to see the exact layout.

Will merge into master after tests run

MartinSenne:
Great! So the first idea of having `Extra` with all supplementary details finally won :)

Will do a PR with some polishing (scaladoc and readme) tomorrow.

lihaoyi:
Yep! Thanks for exploring the design space with me =)

PR #63, opened 2015-11-25
fix dependency logic to work in Scala 2.12 community build

SethTisue:
after https://github.com/lihaoyi/utest/pull/80 was merged,
fastparse was able to run, revealing that the same change
was needed her


lihaoyi:
Looks good to me

PR #64, opened 2015-11-26
Polish `Extra` restructuring.

MartinSenne:
Polishing 

- Added Success and Failure to API
- Made Extra a regular class (instead of case class).
- Corrected / added documentation
- Revised entry in changelog (in readme) and added name.

### DIFF
--- a/fastparse/shared/src/main/scala/fastparse/Api.scala
+++ b/fastparse/shared/src/main/scala/fastparse/Api.scala
@@ -12,6 +12,8 @@ trait Api{
 
   val Result = core.Result
   type Result[+T] = core.Result[T]
+  val Failure = core.Result.Failure
+  val Success = core.Result.Success
 
   val Pass = parsers.Terminals.Pass
   val Fail = parsers.Terminals.Fail
--- a/fastparse/shared/src/main/scala/fastparse/Api.scala
+++ b/fastparse/shared/src/main/scala/fastparse/Api.scala
@@ -11,7 +11,7 @@ trait Api{
   implicit def enclosingFunctionName: Utils.FuncName = macro MacroUtils.impl
 
   val Parsed = core.Parsed
-  type Result[+T] = core.Parsed.Result[T]
+  type Parsed[+T] = core.Parsed[T]
   
   val Pass = parsers.Terminals.Pass
   val Fail = parsers.Terminals.Fail
--- a/fastparse/shared/src/main/scala/fastparse/Api.scala
+++ b/fastparse/shared/src/main/scala/fastparse/Api.scala
@@ -10,11 +10,9 @@ import acyclic.file
 trait Api{
   implicit def enclosingFunctionName: Utils.FuncName = macro MacroUtils.impl
 
-  val Result = core.Result
-  type Result[+T] = core.Result[T]
-  val Failure = core.Result.Failure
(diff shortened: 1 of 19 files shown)


fastparse/shared/src/main/scala/fastparse/Api.scala
```diff
@@ -12,6 +12,8 @@ trait Api{
 
   val Result = core.Result
   type Result[+T] = core.Result[T]
+  val Failure = core.Result.Failure
+  val Success = core.Result.Success
```
lihaoyi: I somewhat-intentionally left these out, because I Want them to be called via `Result.*` rather than just `*`. Otherwise they tend to collide with lots of things called `Success` or `Failure`, e.g. `Try` in the standard library =/ We could name them `ParseSuccess` or `ParseFailure` but I'm more inclined to just name them `Result.Success` and `Result.Failure`

lihaoyi: Perhaps we could renamed `Result` into `Parsed` for even better disambiguation? `Parsed.Success` and `Parsed.Failure` seems just about perfect

lihaoyi:
That's all for now. 

At first I was hesitant with `line`/`col` going into `Extra`, but I guess it makes sense if you want people to pattern match on `Failure` and retrieve them conveniently.

fastparse/shared/src/main/scala/fastparse/Api.scala
MartinSenne: Ok, the name-clash with scala.util.[Try|Success|Failure] is to be avoided. So, `Result` -> `Parsed` is fine. Second, we follow the terminology of having `ed` names for results (e.g.`traced` in Extra).

MartinSenne:
Alright, will adapt the PR by today.

MartinSenne:
Done! Hope you like it!

lihaoyi:
No need for `Parsed.Result`, just `Parsed` will suffice for the trait together with `Parsed.Success` and `Parsed.Failure`

MartinSenne:
IMHO, even though there is no direct need for `Parsed.Result`, it "joins" all Result types, `Parsed.Result`, `Parsed.Success` and `Parsed.Failure`, nicely at a common place. Without reading docs, I would not expect `Parsed` to be the supertype of `Parsed.Success` or `Parse.Failure`.

lihaoyi:
I say we should go with `Parsed` instead of `Parsed.Result`. This fits the pattern that was there before, and even if you didn't know what `Parsed` was it should take <5s to find out.

If you want to debate the convention of putting instances/subtypes in the companion we can do that in gitter, but for now I'll say it's very common: people put implicit instances there due to the implicit search path, and when you have a non-trivial structure like [this](https://github.com/lihaoyi/Scalatex/blob/master/api/src/main/scala/scalatex/stages/Parser.scala#L144) it really does help keep things neat over trying to flatten everything out into the same namespace.

MartinSenne:
Didn't want to start a debate: Simply two different "approaches" of modelling ;)

I reverted the movement of `Result` into `Parsed.Result` and rechecked documentation once again.

PR is thus ready to be merged.

lihaoyi:
image

MartinSenne:
:)

Issue #66, opened 2015-11-30
Parser.rep does not handle max=0 correctly.


MartinSenne:
## Description and Reproduction

When using `.rep` with min = 0 and max = 0

    println( P("  ".rep(min=0, max=0) ~ End).parse("  "))
    println( P("  ".rep(min=1, max=1) ~ End).parse("    "))

the result is

    Success((),2)
    Failure(End:1:3 ..."  ")

Second parser and second result are **correct**.
First parser with min=0 and max=0 and a **`Success`** result is **not correct**, because
* max is not fulfilled (with a successful parse, the "determined" number of repetitions is 1, which is larger than a max of 0)

### Proof (of consistency)

    println( P("" ~ End).parse("  ")) 

delivers

    Failure(End:1:1 ..."  ")

as expected.

## Expected result after fix

* First parser should return a `Failure`. 
* No input is consumed

MartinSenne:
Closed with merge of #67.

PR #67, opened 2015-12-02
#66 Bugfix.

MartinSenne:
Please recheck, if last parameter `false` in

    success(cfg.success, ev.result(ev.initial), index, List.empty[Parser[_]], false)

is ok!

Tests (plus additional tests for bugfix) run through.


lihaoyi:
Looks good to me, thanks!

Issue #77, opened 2016-02-29
Infinite loop while constructing ParseError


jeroenr:
I have this problem where I'm getting stuck in an infinite loop while constructing ParseError. So the actual parsing completes, results into an error and then when I try to construct ParseError it's getting stuck into a loop while building the error trace. I'm using Fastparse version 0.3.4.

It looks like it's getting stuck [here](https://github.com/lihaoyi/fastparse/blob/master/fastparse/shared/src/main/scala/fastparse/parsers/Combinators.scala#L439-L444). The stack looks like this while it's stuck:

```
	  at scala.collection.generic.Growable$class.loop$1(Growable.scala:54)
	  at scala.collection.generic.Growable$class.$plus$plus$eq(Growable.scala:57)
	  at scala.collection.mutable.ListBuffer.$plus$plus$eq(ListBuffer.scala:183)
	  at scala.collection.immutable.List.$colon$colon$colon(List.scala:128)
	  at fastparse.parsers.Combinators$Either.rec$4(Combinators.scala:444)
	  at fastparse.parsers.Combinators$Either.parseRec(Combinators.scala:447)
	  at fastparse.parsers.Combinators$Rule.res$lzycompute$1(Combinators.scala:142)
	  - locked <0x961> (a fastparse.parsers.Combinators$Rule)
	  at fastparse.parsers.Combinators$Rule.fastparse$parsers$Combinators$Rule$$res$1(Combinators.scala:142)
	  at fastparse.parsers.Combinators$Rule.parseRec(Combinators.scala:147)
	  at fastparse.parsers.Combinators$Either.rec$4(Combinators.scala:439)
	  at fastparse.parsers.Combinators$Either.parseRec(Combinators.scala:447)
	  at fastparse.parsers.Combinators$Capturing.parseRec(Combinators.scala:22)
	  at fastparse.parsers.Combinators$Rule.res$lzycompute$1(Combinators.scala:142)
	  - locked <0x964> (a fastparse.parsers.Combinators$Rule)
	  at fastparse.parsers.Combinators$Rule.fastparse$parsers$Combinators$Rule$$res$1(Combinators.scala:142)
	  at fastparse.parsers.Combinators$Rule.parseRec(Combinators.scala:147)
	  at fastparse.parsers.Combinators$Either.rec$4(Combinators.scala:439)
	  at fastparse.parsers.Combinators$Either.parseRec(Combinators.scala:447)
[...]

PR #79, opened 2016-03-02
expose #77 in test & fix by deduplicating traceParsers on every pass

jeroenr:
Without the changes to **fastparse/shared/src/main/scala/fastparse/WhitespaceApi.scala** the added test will blow up the stack


jeroenr:
The build on Travis fails because the master build is unstable.

PR #80, opened 2016-03-07
Added a StringInIgnoreCase parser

jimdickinson:
We're replacing the parser combinators library with fastparse and one thing we're working around is parsing for a set of case insensitive strings. This PR adds that feature with a small change to the StringIn parser, and adds a couple tests.


jimdickinson:
@lihaoyi any thoughts about this? This is to avoid `P( IgnoreCase(str1) | IgnoreCase(str2) | IgnoreCase(str3) | IgnoreCase(str4) )`, which is discouraged in the docs.

jimdickinson:
@lihaoyi any feedback?

jimdickinson:
@lihaoyi any feedback? I don't want to maintain a fork of the library and I think this is a generally useful thing. We really like using fastparse at compstak.com and this is the only missing thing we found.

PR #83, opened 2016-03-27
Add a `to` method to the Api

liff:
Here's a proposed new method to the Api.

The `to` method is an alternative for `map` that in some cases provides slightly more convenient syntax.

Link to the very short [gitter discussion](https://gitter.im/lihaoyi/fastparse?at=56f7d6f48f5147e119f0df85).

Not sure about the name but this is just a proposal at this point.

TODO:
- [ ] agree on the name and
- [ ] write documentation.

### DIFF
--- a/fastparse/shared/src/main/scala/fastparse/Api.scala
+++ b/fastparse/shared/src/main/scala/fastparse/Api.scala
@@ -13,7 +13,8 @@ import fastparse.Utils.HexUtils
 
 class Api[ElemType, Repr]()(implicit elemSetHelper: ElemSetHelper[ElemType],
                             elemFormatter: ElemTypeFormatter[ElemType],
-                            ordering: Ordering[ElemType]) extends ToArityMethods {
+                            ordering: Ordering[ElemType])
+  extends ToArityMethods[ElemType, Repr] {
   type Parsed[+T] = core.Parsed[T, ElemType]
   object Parsed {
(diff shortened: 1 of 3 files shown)


lihaoyi:
As discussed in gitter, the name `.to` is fine, and we want this to go in for 0.4.0. Some work items before then:

- Rebased on master

- Use of the small example snippet inside the doc-site, probably under the section for `Map`, which explains what this is about, when to use it, and it's relationship with `.map`. Adding one example handling `TupleN` for N > 2 would be nice to illustrate the point.

- Port the existing parsers (JSON, CSS, Arithmetic, ...) to use `.to` where appropriate, eliminating calls to `.tupled` on the case-class companions and the nullary `.map(_ => ...)` calls on `P[Unit]`s.

- Explore replacing the implicit classes/conversions with a single method on the `Api` trait, which then takes an implicit parameter e.g. `def to[V](f: T => V)(implicit ToMapper[T, V])`, where `ToMapper` instances can all live in the companion object and not need to be imported. Not sure if this will work, but if it works it means fewer implicits imported into local scope, which is a good thing.

- Possibly add a few (non-example) tests just to validate that it works as expected for `TupleN`s that are not covered by the example tests.

Sorry for the lack of attention so far, let's get this in =)

lihaoyi:
This missed 0.4.0, but I'm still open to getting this in as discussed in gitter. Feel free to ping this issue if there's updates

PR #84, opened 2016-04-11
Added cssparse module

vovapolu:
Moved https://github.com/vovapolu/CssParser here


build.sbt
```diff
@@ -121,11 +121,21 @@ lazy val pythonparse = crossProject.dependsOn(fastparse).settings(
 lazy val pythonparseJVM = pythonparse.jvm
 lazy val pythonparseJS = pythonparse.js
 
+lazy val cssparse = crossProject.dependsOn(fastparse).settings(
```
lihaoyi: Could you format this as:

```scala
lazy val cssparse = crossProject
  .dependsOn(fastparse)
  .settings(name := "cssparse")
  .settings(shared_:*)
```

No need `jvmSettings` if it's just empty

cssparse/shared/src/main/scala/cssparse/CssParser.scala
```diff
@@ -0,0 +1,138 @@
...
+  val substring_match_token = P( "*=" ) map {_ => Ast.SubstringMatchToken}
+  val column_token = P( "||" ) map {_ => Ast.ColumnToken}
+  val CDO_token = P( "<!--" ) map {_ => Ast.CdoToken}
+  val CDC_token = P( "-->" ) map {_ => Ast.CdcToken}
+
+  val delim_token = P( CharIn("#$*+,-./:;<>@^~=!").! ) map Ast.DelimToken
+
+  // any token except function_token
+  val simple_token: Parser[Option[Ast.SimpleToken]] = P(
+    whitespace_token | percentage_token |
```
lihaoyi: Can you align this nicely in 2 or 3 columns? right now it's pretty

cssparse/shared/src/main/scala/cssparse/CssParser.scala
```diff
@@ -0,0 +1,138 @@
...
+
+  val ident_token = P( ("-".? ~ (CharIn('a' to 'z', 'A' to 'Z', "_") | escape) ~
+    (CharIn('a' to 'z', 'A' to 'Z', '0' to '9', "_-") | escape).rep).! ) map Ast.IdentToken
+
+  val function_token = P( ident_token.! ~ "(" ) map Ast.FunctionToken
+
+  val at_word_token = P( "@" ~ ident_token.! ) map Ast.AtWordToken
+
+  val hash_token = P( "#" ~
+    (CharIn('a' to 'z', 'A' to 'Z', '0' to '9', "_-") | escape).rep(1).! ) map Ast.HashWordToken
```
lihaoyi: As a convention use the `.map(...)` syntax to be consistent with the rest of the code in the repo

cssparse/shared/src/test/scala/cssparse/PrettyPrinterTests.scala
```diff
@@ -0,0 +1,332 @@
+package cssparse
+
+import utest._
+
+object PrettyPrinterTests extends TestSuite {
+  val tests = this {
+    'basic {
+      'test1 {
+        val input =
```
lihaoyi: Could you add another test input, which is basically the same thing as `input` but fed into https://cssminifier.com/, and have an assert to make sure that both outputs parse and match the `output` expected

cssparse/shared/src/main/scala/cssparse/Ast.scala
```diff
@@ -0,0 +1,65 @@
+package cssparse
+
+// According to https://www.w3.org/TR/css-syntax-3/
+
+object Ast {
+
+  abstract class ComponentValue
```
lihaoyi: All of these should be `sealed` to make sure people don't accidentally inherit from them, and to send a message to future programmers that this is a "sealed trait hierarchy"

cssparse/shared/src/test/scala/cssparse/PrettyPrinterTests.scala
```diff
@@ -0,0 +1,332 @@
...
+            |  -webkit-text-size-adjust : 100% ;
+            |  -ms-text-size-adjust : 100% ;
+            |}
+            |
+            |body{
+            |  margin : 0 ;
+            |}
+            |
+            |article , aside , details , figcaption , figure , footer , header , hgroup , main , menu , nav , section , summary{
+            |  display : block ;
```
lihaoyi: Could you tweak the pretty-printer to have no space before each `,`, `:` or `;`?

cssparse/shared/src/test/scala/cssparse/PrettyPrinterTests.scala
```diff
@@ -0,0 +1,332 @@
...
+            |}
+            |
+            |input [type = 'search']{
+            |  -webkit-box-sizing : content-box ;
+            |  -moz-box-sizing : content-box ;
+            |  box-sizing : content-box ;
+            |  -webkit-appearance : textfield ;
+            |}
+            |
+            |input [type = 'search']: : -webkit-search-cancel-button , input [type = 'search']: : -webkit-search-decoration{
```
lihaoyi: This should probably be `::` instead of `: : `

cssparse/shared/src/test/scala/cssparse/PrettyPrinterTests.scala
```diff
@@ -0,0 +1,332 @@
...
+            |}
+            |
+            |input [type = 'checkbox'], input [type = 'radio']{
+            |  -webkit-box-sizing : border-box ;
+            |  -moz-box-sizing : border-box ;
+            |  box-sizing : border-box ;
+            |  padding : 0 ;
+            |}
+            |
+            |input [type = 'number']: : -webkit-inner-spin-button , input [type = 'number']: : -webkit-outer-spin-button{
```
lihaoyi: Let's have no spaces around the `=`

cssparse/shared/src/test/scala/cssparse/PrettyPrinterTests.scala
```diff
@@ -0,0 +1,332 @@
...
+            |}
+            |
+            |@media (min-width : 768px){
+            |  . lead {
+            |    font-size : 21px ;
+            |  }
+            |}
+            |
+            |@media screen and (max-width : 767px){
+            |  . table-responsive {
```
lihaoyi: Pretty sure we need to have no space after the `.` when selecting on a css-class

lihaoyi:
The pretty-printer breaks the CSS (e.g. spaces around `.`s, `:`s). Apart from fixing that, we could expand our test suite in various ways to try to catch this sort of bug without too much work:

- You've checked that `prettyprint(input) == output`, but you could also check that `prettyprint(output) == output`. I'm not sure if it would fail to parse right now due to the broken CSS in `output`. It might, because the parser is currently not very strict, but if it fails that will help us catch this sort of broken CSS in output.

- We can use a third-party parser to validate our `output` to make sure it's not doing anything crazy. For example, we could use `CSSParser`:
```scala
haoyi-Ammonite@ load.ivy("net.sourceforge.cssparser" % "cssparser" % "0.9.18")
haoyi-Ammonite@ import com.steadystate.css.parser._
haoyi-Ammonite@ import java.io.StringReader
haoyi-Ammonite@ import org.w3c.css.sac.InputSource
haoyi-Ammonite@ val source = new InputSource(new StringReader("div . h1 { background: #ffcc44; }"));
source: InputSource = org.w3c.css.sac.InputSource@333afc53

haoyi-Ammonite@ val parser = new CSSOMParser(new SACParserCSS3());
parser: CSSOMParser = com.steadystate.css.parser.CSSOMParser@7be3baf2

haoyi-Ammonite@ import org.w3c.css.sac._
import org.w3c.css.sac._
haoyi-Ammonite@ parser.setErrorHandler(new ErrorHandler{
                  def error(ex: CSSParseException) = println("ERROR" + ex)
                  def fatalError(ex: CSSParseException) = println("FATAL ERROR" + ex)
                  def warning(ex: CSSParseException) = println("WARNING" + ex)
                })

haoyi-Ammonite@ val sheet = parser.parseStyleSheet(source, null, null);
ERRORorg.w3c.css.sac.CSSParseException: Error in class selector. (Invalid token " ". Was expecting: <IDENT>.)
WARNINGorg.w3c.css.sac.CSSParseException: Ignoring the whole rule.
sheet: org.w3c.dom.css.CSSStyleSheet =
haoyi-Ammonite@
```

This will help us ensure sanity; not just that we can parse our own input, and parse our own output, but *also* that the stuff that we're spitting out as output is valid to third-party CSS parsers.

- Since we have a pretty-printer that allows us to round-trip things, and we have a third party parser which should let us validate that things are sane, let's make all the `ProjectTests` do the same round-trip validation and third-party-parser validation on the pretty-printed output. In theory, they should all pass and  it shouldn't be much work to write the tests

lihaoyi:
Lastly, the travis build doesn't seem to be passing your new test. No idea why it's behaving differently from running locally but we should fix it before landing

lihaoyi:
That is all for now. In general the parser is in good shape, but the pretty-printer could be improved and the test suite could be made more rigorous before landing.

The output AST is a bit low-level for my liking, but it will do for now, accomplishes the mission I gave you at the outset, and is a big step towards where I want it to go. Let's get this diff into shape and land it into master before we talk about what to do next =)

lihaoyi:
bump @vovapolu

vovapolu:
@lihaoyi I fixed codestyle in parser class and made PrettyPrinter print more clear css. I also added that check `prettyprint(output) == output`, but validation with the third-party parser isn't working right now. The main reason is that my parser can't parse expressions like `tr:last-child` or `!important` into paticular blocks, but only into the sequence of tokens, so PrettyPrinter can't distinguish them from the other css stuctures and print them in the wrong way (`tr: last-child`, `! important`).

lihaoyi:
Sorry for the slow response!

How hard would it be to make the pretty-printer smart enough to identify blocks and other css structures to be able to print them properly e.g. `tr:last-child` or `!important`? Presumably you could run some kind of state-machine over the sequence of tokens to identify these structures and build them into a slightly-richer AST before pretty-printing?

vovapolu:
Yes, I thought about postprocessing of these structures. I think it isn't very hard to implement, and the only difficulty will be the great number of rules that CSS format has. I will try to add several rules soon :)

lihaoyi:
Cool sounds like a plan! 

It's ok if you don't have 100% coverage of all-possible-css-syntaxes out of the box, but let's aim to fully support all of the 4 sample files that you picked, and make sure we can round-trip those through the parser/pretty-printer without corrupting them. If we find we want to add support for less common syntaxes, we can do so later.

cssparse/shared/src/main/scala/cssparse/PrettyPrinter.scala
```diff
@@ -0,0 +1,137 @@
...
+
+object PrettyPrinter {
+
+  val indentSize = 2
+  def indentation(indent: Int, isIndentation: Boolean): String =
+    if (isIndentation) "\n" + " " * indentSize * indent else " "
+
+  def printToken(token: SimpleToken): String = {
+    token match {
+      case IdentToken("and") => "and "
```
lihaoyi: Should there be a space inside the string before the `and`?

vovapolu: There is a space between words in `printComponentValues` method.

cssparse/shared/src/main/scala/cssparse/CssParser.scala
```diff
@@ -0,0 +1,205 @@
...
+
+  val dimension_token = P( number_token.! ~ ident_token.! ) map
+    { case (number, ident) => Ast.DimensionToken(number, ident)}
+
+  val percentage_token = P( number_token.! ~ "%" ).map(Ast.PercentageToken)
+
+  val unicode_range_token = P( CharIn("Uu") ~ "+" ~
+    ((hex_digit.rep(min=1, max=6) ~ "?".rep(min=1, max=5)).! |
+    (hex_digit.rep(min=1, max=6).! ~ "-" ~ hex_digit.rep(min=1, max=6).!) |
+    hex_digit.rep(min=1, max=6).!)).map({
```
lihaoyi: Remove the redundant outer ()s here and in all the other .map calls

cssparse/shared/src/main/scala/cssparse/CssParser.scala
```diff
@@ -0,0 +1,205 @@
...
+  val hex_digit = P( CharIn('0' to '9', 'a' to 'f', 'A' to 'F') )
+
+  val escape = P( "\\" ~ ((!(newline | hex_digit) ~ AnyChar) |
+    (hex_digit.rep(min=1, max=6) ~ whitespace.?)) )
+
+  val whitespace_token = P( comment | whitespace.rep(1) )
+
+  val ws = P( whitespace_token.rep )
+
+  val ident_token = P( ("-".? ~ (CharIn('a' to 'z', 'A' to 'Z', "_") | escape) ~
```
lihaoyi: Break long lines like this into multiple sub-parsers, combined in one final parser, to keep each parser to one line if possible. The only multi-line parsers should have be relatively simple (e.g. A bunch of named parsers separated by `|`s) and not a complicated mix of `|`s and `~`s. Scope the sub-parsers within this `val` to keep them local

cssparse/shared/src/main/scala/cssparse/CssParser.scala
```diff
@@ -0,0 +1,205 @@
...
+  val curlyBracketsBlock =  P( "{" ~ componentValue.rep ~ "}" ).map(values => Ast.CurlyBracketsBlock(values.flatten))
+  val squareBracketsBlock = P( "[" ~ componentValue.rep ~ "]" ).map(values => Ast.SquareBracketsBlock(values.flatten))
+
+  val functionBlock = P( function_token ~ componentValue.rep ~ ")").map({
+    case (Ast.FunctionToken(name), values: Seq[Option[Ast.ComponentValue]]) =>
+      Ast.FunctionBlock(name, Ast.BracketsBlock(values.flatten))
+  })
+
+  val componentValue: Parser[Option[Ast.ComponentValue]] = P( simple_token | bracketsBlock |
+    curlyBracketsBlock | squareBracketsBlock | functionBlock ).map({
```
lihaoyi: Avoid pattern matching on `Any` like this. Call .map on the individual parsers to convert them to the same type before `|`ing them together

cssparse/shared/src/main/scala/cssparse/CssParser.scala
```diff
@@ -0,0 +1,205 @@
...
+    }))
+  })
+
+  val partSelector = P( allSelector | attributeSelector | elementSelector )
+
+  val classSelectorPart = P( "." ~  partSelector ).map(Ast.ClassSelectorPart)
+
+  val pseudoSelectorPart = P( (("::" | ":") ~ ident_token).! ~ ("(" ~ componentValue.rep(1) ~ ")").? ).map({
+    case (name, optValues) =>
+      Ast.PseudoSelectorPart(name, optValues.map(cvs => cvs.flatMap(cv => cv)).getOrElse(ArrayBuffer()))
```
lihaoyi: optValues.toSeq.flatten.flatten would be a lot shorter

lihaoyi:
I'm seeing some errors in the logs:

```
-----------------------------------------------------------------------------------------------------------------------ERROR org.w3c.css.sac.CSSParseException: Error in expression; ':' found after identifier "progid". Line: 3899 Column:17
---------ERROR org.w3c.css.sac.CSSParseException: Error in expression; ':' found after identifier "progid". Line: 6839 Column:17
ERROR org.w3c.css.sac.CSSParseException: Error in expression; ':' found after identifier "progid". Line: 6850 Column:17

```

But it's not making tests fail. Should we be worried?

cssparse/shared/src/main/scala/cssparse/CssParser.scala
```diff
@@ -0,0 +1,205 @@
...
+  val complexSelectorPart = P( pseudoSelectorPart | classSelectorPart )
+
+  val complexSelector = P( partSelector.? ~ complexSelectorPart.rep(1) ).map({
+    case (part, parts) => Ast.ComplexSelector(part, parts)
+  })
+
+  val singleSelector: Parser[Ast.SingleSelector] = P( complexSelector | partSelector | idSelector | allSelector )
+
+  val selectorDelim = P( (ws ~ CharIn(",>+~").! ~ ws) | whitespace_token.rep(1).! ).map({
+    case s if s.startsWith(" ") => " "
```
lihaoyi: Why are we converting multiple white spaces into a single space during parsing? If we need it for pretty-printing we should do this during pretty-printing

vovapolu:
Now errors from validatioin don't make tests fail, but these errors are from lines like this `progid:DXImageTransform.Microsoft.gradient(enabled = false)` which seem wrong to the validation algorithm

cssparse/shared/src/test/scala/cssparse/PrettyPrinterTests.scala
```diff
@@ -0,0 +1,347 @@
...
+            |  from {
+            |    background-position: 40px 0;
+            |  }
+            |  to {
+            |    background-position: 0 0;
+            |  }
+            |}
+          """.stripMargin
+
+        assert(PrettyPrinter.printRuleList(CssRulesParser.ruleList.parse(input).get.value).trim == output.trim)
```
lihaoyi: Can you extract this and the two idtical snippets below into a helper function?

cssparse/shared/src/main/scala/cssparse/CssParser.scala
```diff
@@ -0,0 +1,205 @@
...
+    delim_token ).map({
+      case st: Ast.SimpleToken => Some(st)
+      case _ => None
+    })
+
+  val bracketsBlock =       P( "(" ~ componentValue.rep ~ ")" ).map(values => Ast.BracketsBlock(values.flatten))
+  val curlyBracketsBlock =  P( "{" ~ componentValue.rep ~ "}" ).map(values => Ast.CurlyBracketsBlock(values.flatten))
+  val squareBracketsBlock = P( "[" ~ componentValue.rep ~ "]" ).map(values => Ast.SquareBracketsBlock(values.flatten))
+
+  val functionBlock = P( function_token ~ componentValue.rep ~ ")").map({
```
lihaoyi: Why are some of the parsers camelCase and some snake_case? Should we standardize it?

cssparse/shared/src/main/scala/cssparse/CssParser.scala
vovapolu: Because it's a selector delimiter, and if wrote `ws ~ whitespace_token.rep(1) ~ ws` or just `ws` it wouldn't work since `ws` can mean any number of spaces including 0.

cssparse/shared/src/main/scala/cssparse/CssParser.scala
vovapolu: The idea was that parsers for tokens are snake_case and more abstract parsers are camelCase. But it's just my imagination :) I will change all parsers to the camelCase.

lihaoyi:
That's all for now. After we settle these issues we can merge it and any future improvements can come later

lihaoyi:
can we make validation errors make tests fail if they occur in the prettyprinted css but not in the original css? that will allow the validator to have whatever bugs it wants, and the input css to have whatever bugs it wants, while still ensuring we ourselves arent introducing problems with the css

lihaoyi:
Looks good. Errors look like a change in one of the test repos which we havent version-pinned. Thanks  @vovapolu!

PR #87, opened 2016-05-04
changed https to http in documentation link

krux02:
My browser told be that the https connection is secure and I could not enter the documentation through that link, so I changed it to an http link that works nicely. I think having http instead of https for documentation is something most people can live with.


krux02:
the failed checks are not because of this pull request.

lihaoyi:
Thanks!

PR #93, opened 2016-06-05
[WIP] Two new type-parameters for element type of input in the parser and representaion of the output

vovapolu:
This pull-request adds some new features to the fastparse:
- `String` was replaced by more abstract `IndexedSeq` with new type-parameter `ElemType` in the core. It allows to create parsers for collections of arbitary classes. 
- There is also a new type-parameter 'Repr' that means representation of the result in capturing, e.g. `IndexedSeq[ElemType]` is casted to the `Repr` when we want to retrieve the parser results. 
- So the string parsing api now is just a special case with `ElemType` === `Char` and `Rerp` === `String`
- New api for the byte parsing (`ElemType` === `Byte` and `Rerp` === `Array[Byte]`)
- Tests covering basic cases in byte parsing by analogy with fastparse basic tests.
- New parser for BMP format with tests taken from [wiki](https://en.wikipedia.org/wiki/BMP_file_format) and tests of parsing the real bmp file.
- New byte tests and bmp parsing were put to the separate module `byteparse`.
- Perfomance benchmarks for all the current modules. E.g. `PerfTests` from scalaparse was moved to new module `perftests`, where were also added new benchmarks. All of them uses the same idea of measuring how many times each parser can process some big file in limited time. 

Adding new type-parameters and transforming native `String` to the `IndexedSeq` structure, unfortunately, brings perfomance reduction in 10-20% approximately.


cssparse/jvm/src/test/scala/cssparse/TestUtil.scala
```diff
@@ -1,6 +1,6 @@
 package cssparse
 
-import fastparse.all._
+import fastparse.allString._
```
lihaoyi: Let's leave this as `fastparse.all._` (and the other as `fastparse.noApi._`) for the time being. That's not to say we can't change it, but this change is already going to be huge enough =) 

Keeping the peripheral stuff backwards compatible and the use-site code unchanged lets us focus on the core changes that really matter. Once this huge diff lands, we'll have plenty of time to go back and tweak these things

fastparse/shared/src/main/scala/fastparse/ParserHelper.scala
```diff
@@ -0,0 +1,46 @@
+package fastparse
+
+import fastparse.Utils.HexUtils
+
+import scala.collection.mutable.ArrayBuffer
+
+trait ParserHelper[ElemType] {
+  val delimiter: ElemType
+  val emptyElem: IndexedSeq[ElemType]
+  def convertToString(input: IndexedSeq[ElemType]): String
```
lihaoyi: Let's call this `prettyprint` since that's what AFAICT it does

scalaparse/shared/src/test/scala/scalaparse/unit/FailureTests.scala
```diff
@@ -123,7 +123,7 @@ object FailureTests extends TestSuite{
         |  1
         |}
       """.stripMargin,
-      expected = """ ";" | Newline.rep(1) | End """,
+      expected = """ ";" | Newline.rep(1) | End() """,
```
lihaoyi: Can we give `End` and all the other no-arg parsers `toString` methods without the `()`, for backwards-compatiblity for now?

utils/shared/src/main/scala/fastparse/Utils.scala
```diff
@@ -82,17 +105,40 @@ object Utils {
       }
       res.mkString
     }
-    def compute(chars: Seq[Char]) = {
-      val first = chars.min
-      val last = chars.max
+  }
+
+  trait ElemHelper[Elem] {
```
lihaoyi: Possible to move this into the same file as all the `ParserHelper` stuff? I'd like to keep all the type-specific helpers in one file so we can quickly skim over them and see what's there

fastparse/shared/src/main/scala/fastparse/ParserHelper.scala
```diff
@@ -0,0 +1,46 @@
+package fastparse
+
+import fastparse.Utils.HexUtils
+
+import scala.collection.mutable.ArrayBuffer
+
+trait ParserHelper[ElemType] {
```
lihaoyi: We should avoid calling things `Helper`, since it doesn't tell us anything about what it's used for (all code is meant to "help" people...)

Could this be called `ElemTypeMessageFormatter`? Or something else? I'm not entirely sure what would be the best name, but `ParserHelper` is probably too vague

fastparse/shared/src/main/scala/fastparse/ParserHelper.scala
```diff
@@ -0,0 +1,46 @@
...
+  val delimiter: ElemType
+  val emptyElem: IndexedSeq[ElemType]
+  def convertToString(input: IndexedSeq[ElemType]): String
+  def literalize(input: IndexedSeq[ElemType]): String
+}
+
+trait ConversionHelper[ElemType, Repr] {
```
lihaoyi: Perhaps we could call this `ResultBuilder` or something?

fastparse/shared/src/main/scala/fastparse/Api.scala
```diff
@@ -1,51 +1,130 @@
...
+  def ByteSeq[T](bytes: T*)(implicit num: Numeric[T]) = bytes.map(b => num.toInt(b).toByte).toArray
+  def BS[T](bytes: T*)(implicit num: Numeric[T]) = ByteSeq[T](bytes: _*)
+  type ByteSeq = Array[Byte]
+  type BS = ByteSeq
+
+  implicit def wspByteSeq(seq: Array[Byte]): P0 =
+    if (seq.length == 1) parsers.Terminals.ElemLiteral(seq(0))
+    else parsers.Terminals.Literal(seq)
+
+  implicit class ByteSeqUtils(s: String) {
```
lihaoyi: Can we implement these as normal functions instead of extension methods? e.g. `hexChars(s: String)` or `strToByte(s: String)`

lihaoyi:
Looks great! Left some nitpicks, but we can talk more about how it's going later

How is the performance of the original scala-string-parsing benchmarks with the current set of changes?

lihaoyi:
Nice work @vovapolu!

PR #96, opened 2016-06-29
Parser for .class bytecode files

vovapolu:
This pull-request contains parser for bytecode .class files. 
It parses file in two steps:

1. Initial reading and transforming of basic structures describing [here](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.7) (File `ClassParser` constains object 'Info', which describing these structures, and parsers for .class files).

2. Converting these structures into more concise and light classes (There are also object 
`Ast` in file `ClassParser` with more broad system of structures and method `convertToAst` that converts `Info` representation into `Ast`)

  2.1 During this convertion the [attribute](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.7) parsing is performed. (All parsers and methods for this step are situated in file `ClassAttributes`. There is a `Map` with all the attributes associated with its unique names).

  2.2 The last step is code parsing decribed in file `CodeParser`. Each bytecode command has corresponding case class or object which are used in the main method `parseCode`. 

There were also added several tests. The first group of tests is basic checks of correctness of attribute, code and Ast parsing. The second group is `ProjectTests` that downloads existing java repositories, complies them into .class files, does the regular parsing and compares the ouput of built-in `ClassLoader` with this parser.


byteparse/shared/src/main/scala/byteparse/classparse/ClassAttributes.scala
```diff
@@ -0,0 +1,183 @@
...
+
+  case class DeprecatedAttribute() extends Attribute
+
+  case class BootstrapMethod(bootstrapMethodRef: MethodHandle, bootstrapArguments: Seq[PoolItem])
+
+  case class BootstrapMethodsAttribute(bootstrapMethods: Seq[BootstrapMethod]) extends Attribute
+
+  import ByteUtils.BE._
+
+  val attributeParsers = Map[String, (ClassFileInfo) => Parser[Attribute]](
```
lihaoyi: Could we change this from `Map[String, (ClassFileInfo) => Parser[Attribute]]` to a `Map[String, Parser[ClassFileInfo => Attribute]]`? That would allow us to initialize the parser only once, which shouldn't be too slow but isn't particularly fast since FastParse isn't optimized for initialization speed.

lihaoyi: Also, the `Map` here seems big enough it's worth lifting each `ClassFileInfo => Attribute` into a standalone method/factory, and then referring to it in the `Map`. As it stands now the `Map` is kind of sprawling and difficult to skim through, e.g. if we forgot one case it would not be obvious

byteparse/shared/src/main/scala/byteparse/classparse/ClassAttributes.scala
```diff
@@ -0,0 +1,183 @@
...
+        P( AnyWordI ).map(idx =>
+          ConstantValueAttribute(
+            classInfo.getInfoByIndex[PoolInfo](idx).get match {
+              case BasicElemInfo(elem) => elem
+              case StringInfo(idx) => StringElem(classInfo.getStringByIndex(idx))
+            }))
+    },
+    "Code" -> {
+      (classInfo: ClassFileInfo) => {
+        val exceptionHandler =
```
lihaoyi: As a convention, could we rewrite these as

```scala
val exceptionHandler = {
  val start_pc = AnyWordI
  val end_pc = AnyWordI
  val handler_pc = AnyWordI
  val catch_type = AnyWordI.map(idx =>
    if (idx == 0) None
    else Some(Class(classInfo, classInfo.getInfoByIndex[ClassInfo](idx).get))
  )
  P( start_pc ~ end_pc ~ handler_pc ~ catch_type).map(ExceptionHandler.tupled)
}
```

I think it would be much easier to read than having the field-names in the comments, and the final `P( start_pc ~ end_pc ~ handler_pc ~ catch_type).map(ExceptionHandler.tupled)` is much easier to skim.

This applies to all the other `P(...)` blocks as well; if it's getting large and gnarly and spread over 8 lines of code, let's extract all the complex bits into local `val`s before and keep the `P(...)` block to 1, maybe 2 lines if possible

byteparse/shared/src/test/scala/byteparse/classparse/ClassTests.scala
```diff
@@ -0,0 +1,1180 @@
...
+          "CAkQAQwJPAEQCWgBFAmUARgJtAEcCcQBIAngASQKDAEoCigBLApUATQKdAE8CoABQAqcAUQKvAF" +
+          "ICtQBUAs8AVQLSAFcC2gBYAt0AWQLkAFoC/gBbAwgAXAMQAFsDFgBfAyAAYAMqAGEDNgBiAzkAY" +
+          "AM/AF8DRQBmA3YAZwOQAGgDmABnA54AagOhAGwDvABtA8AAbgPDAG8DxwBwA8oAcQPOAHID0QBz" +
+          "A9UAdgPvAHcD8wB4BGgAegRwAHsEcwB9BHsAfgR+AIAEhgCBBIkAgwSjAIYEpgCHBLYAiAS+AIk" +
+          "ExACKBMoAiwBeAAAA2QAf/wKDAB8HAF8BAQEBAQEEAgMBAQEAAwcAYAcAYQcAYQcAYgcAYgcAYg" +
+          "cAYwcAZAcAZQcAYwcAZAcAZQEBAQEAABEH/AACART8ABwB/AAuAfoAFPwAAgH8AAkBFfoABfoAB" +
+          "f8APAAlBwBfAQEBAQEBBAIDAQEBAAMHAGAHAGEHAGEHAGIHAGIHAGIHAGMHAGQHAGUHAGMHAGQH" +
+          "AGUBAQEBAQEHAGMHAGMBAQAA+AAb/AAdAQYGBvwAAwcAYv4ASgcAYgcAYgEPDwwaCgoK+QAZ/AA" +
+          "OAUEHAGIAAQBmAAAAAgBn").toByteArray
+        /* CodeTest.class from compiled class CodeTest.java */
```
lihaoyi: Could we assert things about the constant pool of this class? We don't need to check every since entry, but we should at least spot-check some of the values to make sure some of them e.g. `100000L 2.71828 "Do you want to add a tab?\tNo Problem!" "\n->Control Structures" 1337` which we can see in the source code are present in the constant pool.

byteparse/shared/src/test/scala/byteparse/classparse/ClassTests.scala
```diff
@@ -0,0 +1,1180 @@
...
+
+        val Parsed.Success(parsedClassInfo, _) = ClassParser.classFile.parse(classFile)
+        val parsedClass = ClassParser.Ast.convertToAst(parsedClassInfo)
+
+        assert(parsedClass.fields == ArrayBuffer(
+          Field(
+            "CONST_VAL",
+            "I",
+            FieldFlags(true, false, false, true, true, false, false, false, false),
+            ArrayBuffer(ConstantValueAttribute(IntElem(42))))))
```
lihaoyi: Could we format the closing parentheses on new lines, similar to the rest of the code e.g. https://github.com/lihaoyi/fastparse/blob/master/pythonparse/shared/src/test/scala/pythonparse/UnitTests.scala#L80-L89 and https://github.com/lihaoyi/fastparse/blob/master/pythonparse/shared/src/test/scala/pythonparse/UnitTests.scala#L310-L315? Good to keep things consistent. Someday scalafmt will automate all this but for now we should do it manually

byteparse/shared/src/main/scala/byteparse/classparse/CodeParser.scala
```diff
@@ -0,0 +1,514 @@
+package byteparse
+package classparse
+
+import fastparse.allByte._
+
+object CodeParser {
+  sealed abstract class OpCode
+
+  case class AALoad() extends OpCode
```
lihaoyi: All the `case class`s which take no arguments we should make into `case object`s, to save allocations and improve perf across the board

byteparse/shared/src/main/scala/byteparse/classparse/ClassParser.scala
```diff
@@ -0,0 +1,456 @@
...
+    P( BS(0xCA, 0xFE, 0xBA, 0xBE) ~
+       AnyWordI /*minor_version*/ ~
+       AnyWordI /*major_version*/ ~
+       AnyWordI /*constant_pool_count*/.flatMap(l =>
+         constantPool(l - 1).map(_.reverse)) ~
+       AnyWord.! /*access_flags*/ ~
+       AnyWordI /*this_class*/ ~
+       AnyWordI /*super_class*/ ~
+       AnyWordI /*interfaces_count*/.flatMap(l =>
+         P( AnyWordI.rep(exactly=l) )) ~
```
lihaoyi: We probably don't need to wrap these inner-parsers in `P(...)`s. `P(...)` doesn't actually change the type of things; the main purpose of `P(...)` is to 

- provide a human-readable marker in the error-stack, and 
- to allow recursion. 

The first one doesn't really apply here (I think it'll just pick up the same `classFile` name as the outer `P`, which is not very helpful) and neither does the second (since there's no recursion here) so I'd just drop it

byteparse/shared/src/main/scala/byteparse/classparse/ClassParser.scala
```diff
@@ -0,0 +1,456 @@
...
+       AnyWordI /*bootstrap_method_attr_index*/ ~
+       AnyWordI /*name_and_type_index*/
+     ).map(InvokeDynamicInfo.tupled)
+
+  val singleConstantPoolItem = P( constantClassInfo      | constantFieldRefInfo |
+                                  constantMethodRefInfo  | constantInterfaceMethodRefInfo |
+                                  constantStringInfo     | constantIntInfo |
+                                  constantFloatInfo      | constantNameAndTypeInfo |
+                                  constantUtf8Info       | constantMethodHandleInfo |
+                                  constantMethodTypeInfo | constantInvokeDynamicInfo )
```
lihaoyi: In general, this is the only sort of parser which I'd say is OK to be spread over several lines: where it's already basically as compact as possible and it's only multi-line because of how many identical cases there are. 

For most other parsers, if it's multi-lined with a bunch of heterogenous different things inside, we should try to extract those into local `val`s and keep the individual parsers single-line if possible

byteparse/shared/src/main/scala/byteparse/classparse/ClassParser.scala
```diff
@@ -0,0 +1,456 @@
...
+
+    object MethodFlags {
+      def apply(fs: Array[Byte]) = new MethodFlags(fs)
+    }
+
+    case class ClassFlags(accPublic: Boolean,     accFinal: Boolean,    accSuper: Boolean,
+                          accInterface: Boolean,  accAbstract: Boolean, accSynthetic: Boolean,
+                          accAnnotation: Boolean, accEnum: Boolean){
+
+      def this(fs: Array[Byte]) = this(getFlag(fs(1), 0x01), getFlag(fs(1), 0x10), getFlag(fs(1), 0x20),
```
lihaoyi: Let's format this function call, and others like it, as 

```scala
def this(fs: Array[Byte]) = this(
  getFlag(fs(1), 0x01), getFlag(fs(1), 0x10), getFlag(fs(1), 0x20),
  getFlag(fs(0), 0x02), getFlag(fs(0), 0x04), getFlag(fs(0), 0x10),
  getFlag(fs(0), 0x20), getFlag(fs(0), 0x40)
)
```

The current convention in the codebase is 

- parameter lists at definition-sites and vertical-aligned with closing-paren on same line, 
- parameter lists at call-sites are indented two-spaces, with closing paren on new line un-indented

It's a bit of an idiosyncratic formatting style which is basically what IntelliJ defaults to, but let's maintain it until we can get something like scalafmt to take over formatting responsibilities.

byteparse/shared/src/main/scala/byteparse/classparse/CodeParser.scala
```diff
@@ -0,0 +1,514 @@
...
+  case class Ret(index: Int) extends OpCode
+  case class Return() extends OpCode
+
+  case class SALoad() extends OpCode
+  case class SAStore() extends OpCode
+  case class SIPush(short: Short) extends OpCode
+
+  case class TableSwitch(defaultOffset: Int, low: Int, high: Int, offsets: Seq[Int]) extends OpCode
+
+  case class Wide() extends OpCode // correct behavior for this command hasn't been implemented yet
```
lihaoyi: Any idea what kind of Java code we'd need to generate a `Wide()` opcode? It's worth mentioning in the PR or in a comment here just so we don't forget it

byteparse/shared/src/test/scala/byteparse/classparse/ClassTests.scala
```diff
@@ -0,0 +1,1180 @@
...
+          "AGUBAQEBAQEHAGMHAGMBAQAA+AAb/AAdAQYGBvwAAwcAYv4ASgcAYgcAYgEPDwwaCgoK+QAZ/AA" +
+          "OAUEHAGIAAQBmAAAAAgBn").toByteArray
+        /* CodeTest.class from compiled class CodeTest.java */
+
+
+        val Parsed.Success(parsedClassInfo, _) = ClassParser.classFile.parse(classFile)
+        val parsedClass = ClassParser.Ast.convertToAst(parsedClassInfo)
+
+        assert(parsedClass.methods(1).attributes(0).asInstanceOf[CodeAttribute].code ==
+            ArrayBuffer(
```
lihaoyi: Question: we are asserting that this literal is the output of `.code`, how do we know that this expected output is correct? Did you compare it against some other tool e.g. `javap` to validate that all the bytecodes are the same, and all the indices line up?

byteparse/shared/src/main/scala/byteparse/classparse/CodeParser.scala
vovapolu: I think we need a really huge java file, where size of addresses in the code go beyond 1 or 2 bytes. And since `ProjectTests` passes correctly, it almost never happens as I think.

byteparse/shared/src/main/scala/byteparse/classparse/CodeParser.scala
```diff
@@ -0,0 +1,514 @@
...
+
+  val opCodeParsers = {
+
+    val localIndex = P( AnyByte.! ).map(b => b(0) & 0xff)
+    val poolIndex = P( AnyWordI )
+    val offsetIndex = P( AnyWordI ).map(i => i.toShort)
+    val offsetIndexWide = P( AnyDwordI ).map(i => i.toInt) // TODO It's done because Dword is actually unsigned
+
+    Map[Int, Parser[OpCode]](
+      0x32 -> Pass.map(_ => AALoad()),
```
vovapolu: I think we just should add new Terminal for this. For instance, `PassWith`.

byteparse/shared/src/main/scala/byteparse/classparse/ClassAttributes.scala
lihaoyi: Bump @vovapolu, AFAICT this hasn't happened yet

byteparse/shared/src/test/scala/byteparse/classparse/ClassTests.scala
```diff
@@ -0,0 +1,1482 @@
...
+          BasicElemPool(StringElem("(Ljava/lang/Object;)Z")),
+          BasicElemPool(StringElem("java/lang/Integer")),
+          BasicElemPool(StringElem("parseInt")),
+          BasicElemPool(StringElem("(Ljava/lang/String;)I")),
+          BasicElemPool(StringElem("(I)Ljava/lang/String;"))
+        ))
+
+        assert(parsedClass.methods(1).attributes(0).asInstanceOf[CodeAttribute].code ==
+            ArrayBuffer(
+              GetStatic(2),          //   0: getstatic     #2                  // Field java/lang/System.out:Ljava/io/PrintStream;
```
lihaoyi: Could we add a comment saying that the comments on the right are coming from `javap`? Otherwise it's not clear whether they're the output of a third-party parser or whether they're the output of the byteparse parser reformatted into a third-party-parser's syntax

vovapolu: It's at the end of file

fastparse/shared/src/main/scala/fastparse/parsers/Terminals.scala
```diff
@@ -23,6 +23,10 @@ object Terminals {
     override val toString = "Pass"
   }
 
+  case class PassWith[T, ElemType, R](t: T) extends Parser[T, ElemType, R]{
```
lihaoyi: This should have a scaladoc comment, even if it's a short one-liner. Psychologically it makes a big difference to someone browsing the code or API in their editor, as it's a signal that this is something that you should use. 

In Scala, unlike in Java, libraries tend to be less strict about enforcing the private/public distinction, so the existence of documentation takes the role of indicating whether a function or class is something you should use, or whether it's some internal thing you can touch but shouldn't

lihaoyi:
Some final nits; need some comments, and tweak the formatting of multi-line parsers as described here https://github.com/lihaoyi/fastparse/pull/96#discussion_r70471070.

After that, flesh out the PR description to have a more detailed few-paragraph summary of what this 3000-line PR is about for future generations to skim: summarize why we are doing this, how it was implemented, what changes we made to fastparse/byteparse to accommodate it, what the test strategy is. 

I'm getting on an airplane soonish and won't have reliable internet for the weekend, but I'll be happy to merge once that stuff is done

byteparse/shared/src/test/scala/byteparse/classparse/ClassTests.scala
lihaoyi: Could you move it to the top of the block? People tend to read top-to-bottom and most of the comments are pretty meaningless until you've seen the `output from the javap -c` comment. Perhaps also flesh it out a bit and say "manually run and pasted here for reference, to verify that the parser is doing the right thing" or something like that.

byteparse/shared/src/main/scala/byteparse/classparse/ClassAttributes.scala
vovapolu: What do you mean? I've changed `exceptionHandler` and some other parser definitions.

lihaoyi: Did you push your changes? I'm looking on github and it `exceptionHandler` still looks like
```scala
+  val code = {
 +      val exceptionHandler =
 +        P( AnyWordI /*start_pc*/ ~
 +           AnyWordI /*end_pc*/ ~
 +           AnyWordI /*handler_pc*/ ~
 +           AnyWordI /*catch_type*/
 +        ).map{
 +          case (spc: Int, epc: Int, hpc: Int, ctype: Int) =>
 +            (classInfo: ClassFileInfo) => ExceptionHandler(
 +              spc,
 +              epc,
 +              hpc,
 +              if (ctype == 0) None
 +              else Some(Class(classInfo, classInfo.getInfoByIndex[ClassInfo](ctype).get)))
 +        }
```

And many (most?) of the other parsers look the same, still spread over multiple lines with `/**/` comments on each terminal rather than assigning them to names

vovapolu: Oh :) My mistake. I didn't read it properly. Yeah, I will fix it.

xeno-by:
Do you guys already have a use case for this parser, or it was done just as a case study for byteparse?

vovapolu:
@xeno-by Initially it was conceived as the example of usage of byteparse, but the fact that it's scala-js compatible makes the applications of this parser pretty wide. For instance, combining with [Metascala](https://github.com/lihaoyi/Metascala) it can be used as JS bytecode interpreter.

lihaoyi:
We don't have a current use for it; it's main use is as a demo. However, it
was picked as a project to be potentially usable.

I'm not sure who would want to use it on the JVM since ASM exists and is
vastly more mature and way faster, but maybe someday someone will be
porting scalac or metascala to run on Scala.js, and ASM won't be an option.

Although I don't have any current plans to do so, this parser could act as
a stand-in to get things working until you decide it's worth hand-rolling
your own. Kind of like how Scalaparse is being used by scala-meta.

lihaoyi:
Not why the JDK8 build is misbehaving, but looks good to me.

PR #97, opened 2016-07-21
Iterator parser

vovapolu:
This pull-request adds new type of parsing - `Iterator` parsing that can process streaming data without holding all data in the memory. 

It introduces a new class - `ParserInput` which is able to contain two types of data: `IndexedSeq` and `Iterator` of `IndexedSeq`s. This class now is used as a source for parsing instead of `IndexedSeq` or `String` in previous versions. The most interesting and important method in it is `dropBuffer` that has special behavior for `Iterator` mode, it drops "prefix" of data from inner buffer in `ParserInput`, thus reducing the memory consumption during parsing. The details of how it works are described in the docs of `ParserInput`. 

Another modification is adding new flags into the `ParseCtx` class. They are `isFork`, `isNoCut` and `isCapturing`. They control calling of `dropBuffer` and they are changed when the parsing process goes into corresponding parser unit. 'isFork' turns on in `Either`, `Optional` and `Repeat` parsers, `isNoCut` in `NoCut`, `isCapturing` in `Capturing` and in `WhitespaceApi.CustomSequence` (to completely block any `dropBuffer`). 

### ScalaParse performance test on master (in avg.)
First number is a count of regular parsings of a quite big file during 10 secs, seconds is the same parsings, but with invalid file and tracing of the result `Failure`. 
```
76 15
```

### ScalaParse performance test on this PR (in avg.)
```
72 15 
```

### ScalaParse Iterator tests

The input of ScalaParse was wrapped into the `Iterator` by dividing it in batches with varying size. And using this iterator, as usual, the number of passes within 10 seconds was measured with the maximum size of `ParserInput` inner buffer. 

| Size of batch | Avg. result | Max buffer size |
| --- | --- | --- | 
| 1 | 48 | 1587 |
| 2 | 51 | 1587 |
| 4 | 50 | 1587 |
| 16 | 54 | 1587 |
| 64 | 57 | 1635 |
| 1024 | 58 | 2523 |
| 4096 | 57 | 5595 | 

### Distibutions of buffer size during drops in parsing an iterator data

The 1-sized `Iterator` was chosen for testing distribution of inner buffer size during parsing. 

| Buffer Length | Count |
| --- | --- | --- | 
|0| 5139|
|1| 17564|
|2| 3456|
|3| 1056|
|4| 1525|
|5| 421|
|6| 1404|
|7| 311|
|8| 2382|
|9| 1448|
|10| 1638|
|11-33| 6513|
|34-56| 898|
|57-79| 464|
|80-102| 321|
|103-136| 74|
|138-181| 96|
|185-245| 103|
|246-312| 103|
|315-477| 113|
|485-958| 88|
|996-1587| 22|


byteparse/jvm/src/test/scala/byteparse/LargeBmpIteratorTests.scala
```diff
@@ -0,0 +1,60 @@
...
+import java.net.URL
+
+import fastparse.{IteratorParserInput, LoggedParsedInput}
+
+import utest._
+import fastparse.allByte._
+import BmpParser._
+
+object LargeBmpIteratorTests extends TestSuite {
+  val url = "https://raw.githubusercontent.com/lihaoyi/fastparse/master/byteparse/jvm/src/test/resources/lena.bmp"
```
lihaoyi: Could we read this off disk as an InputStream rather than reading off the web? Otherwise we'll end up in odd situations where the tests will pass on the PR but after you commit it the repo on github changes and the tests fail, or vice versa

byteparse/jvm/src/test/scala/byteparse/LargeBmpIteratorTests.scala
```diff
@@ -0,0 +1,60 @@
...
+  def stream = new Iterator[IndexedSeq[Byte]] {
```
lihaoyi: Is there a performance difference between using `IndexedSeq[Byte]` and `Array[Byte]`?

fastparse/shared/src/main/scala/fastparse/ParserApi.scala
```diff
@@ -24,17 +24,20 @@ trait ParserApi[+T, ElemType, Repr] {
   /**
    * Repeats this parser 0 or more times
    */
-  def rep[R](implicit ev: Repeater[T, R]): Parser[R, ElemType, Repr]
+  def rep[R](implicit ev: Repeater[T, R],
+             formatter: ElemTypeFormatter[ElemType]): Parser[R, ElemType, Repr]
```
lihaoyi: Can the `formatter` be part of the `ParserInput` object, rather than being passed in anywhere as an implicit? I assume it won't change between the various parsers run on the same input, so it would make sense to have it specified once rather than specified all over the place

utils/shared/src/main/scala/fastparse/ParserInput.scala
```diff
@@ -0,0 +1,92 @@
+package fastparse
+
+import scala.collection.mutable.ArrayBuffer
+
+abstract class ParserInputLogger {
+  def logDropBuffer(index: Int)
+  def logApply(index: Int)
+}
```
lihaoyi: I think we can remove this class, and leave it to who-ever wants logging (e.g. our own tests...) to inherit from `ParserInput` and override `dropBuffer` and `apply` themselves to do logging. It doesn't seem like a common enough thing that we need to provide a standalone class for people to work with

fastparse/shared/src/main/scala/fastparse/core/Parsing.scala
```diff
@@ -370,7 +342,32 @@ trait Parser[+T, ElemType, Repr] extends ParserResults[T, ElemType] with Precede
             instrument: (Parser[_, _, _], Int, () => Parsed[_, ElemType]) => Unit = null)
            (implicit formatter: ElemTypeFormatter[ElemType])
             : Parsed[T, ElemType] = {
-    parseRec(new ParseCtx(input, 0, -1, this, index, instrument), index).toResult
+    parseRec(
```
lihaoyi: This and `parseIterator` below should probably just instantiate the relevant `ParserInput` and forward directly to `parseInput` to avoid all the repetition

utils/shared/src/main/scala/fastparse/ParserInput.scala
```diff
@@ -0,0 +1,92 @@
...
+
+abstract class ParserInput[ElemType] {
+  def apply(index: Int): ElemType
+  def dropBuffer(index: Int): Unit
+  def slice(from: Int, until: Int): IndexedSeq[ElemType]
+  def length: Int
+  def innerLength: Int
+  def isReachable(index: Int): Boolean
+}
```
lihaoyi: This should be stable enough to document all these things. Apart from what everything does, some subtleties like `why do we need dropBuffer?` or `what's the difference between length and innerLength?` and `what is isReachable and what do we use it for?` is worth spending extra effort documenting

lihaoyi:
Left some high-level notes of issues we should fix before doing a more detailed in-the-weeds review. 

As part of this review, I'd like you to run some performance benchmarks on various parts of the design space and post the results here, for future reference. For example:

- What perf impact does this diff have on *non-iterator* String/Array[Byte] parsing? We have extra code in some of the main code paths setting flags, after all, and I'd like to know if/how-much it slowed things down

- What perf impact does the various sized chunks in `Iterator[T]` parsing have? e.g. for chunk size 1, 2, 4, 8... 512, 1024, 2048 chars/bytes, on streaming ScalaParse, CssParse, BmpParse and JsonParse? And how does it compare to the non-Iterator parsing performance?

- What is the distribution of lengths of the buffer when `dropBuffer` is called, for ScalaParse, CssParse, BmpParse, ClassParse (?), for the tiniest possible buffer (1 byte/char)?

While tabulating this information is not strictly part of the "code", they would really help us understand and review the actual behavioral characteristics of this change, and will serve as a baseline of facts that future discussions around fastparse can use.

fastparse/shared/src/test/scala/fastparse/iterator/IteratorTests.scala
```diff
@@ -0,0 +1,62 @@
+package fastparse.iterator
+
+import fastparse.all._
+import utest._
+
+object IteratorTests extends TestSuite {
```
lihaoyi: I don't think the unit tests we have here are sufficient; we should add some unit tests that (or modify these to) subclass `ParserInput` and hook into `dropBuffer` in order to verify that `dropBuffer` indeed gets called at exactly the right moments within these tests, and that the buffer does not grow past whatever size we think it should not grow past. 

That will let us be confident that we're actually being as streaming as we should be streaming, and not accidentally loading the whole thing into memory before dealing with it

byteparse/jvm/src/test/scala/byteparse/LargeBmpIteratorTests.scala
vovapolu: What do you mean? `parseIterator` takes `Iterator[IndexedSeq[Byte]]`, so I can't use `Array` directly.

lihaoyi: Could we make `parseIterator` take an `Iterator[Array[Byte]]`? I'm pretty sure `IndexedSeq[Byte]` ends up take 5-20x as much memory as `Array[Byte]` due to boxing converting each byte to a 4-8 byte pointer to an object with a 8-12 byte header with 1 byte of data each...

vovapolu: I can write `Iterator[WrappedArray[Byte]]` which, I think, would be faster than simple `IndexedSeq`, but a little bit slower that `Array`.

perftests/shared/src/main/scala/perftests/Utils.scala
```diff
@@ -12,7 +17,94 @@ object Utils {
...
+                                  (implicit formatter: ElemTypeFormatter[ElemType],
+                                            converter: ResultConverter[ElemType, Repr]): Unit = {
+
+    val results = Utils.benchmark(s"$name Benchmark",
+      Seq(Some(() => parser.parse(data)),
+          dataFailOpt.map(dataFail =>
+            () => parser.parse(dataFail).asInstanceOf[Parsed.Failure[ElemType]].extra.traced)).flatten)
+    println(results.map(_.mkString(" ")).mkString("\n"))
+
+    val sizes = Seq(1, 2,/* 4, 16, 64,*/ 1024/*, 4096*/)
```
lihaoyi: Can we leave all of these un-commented?

fastparse/shared/src/test/scala/fastparse/iterator/IteratorTests.scala
```diff
@@ -0,0 +1,105 @@
...
+    'cuts {
+      'capturing {
+        val p = P( "a" ~/ "b" ~/ "c")
+        val capt = P( p.! ~ p.! ~ p.!)
+        val input = toInput("abcabcabc")
+        val Parsed.Success(res, i) = capt.parseInput(input)
+        assert(
+          i == 9,
+          res == ("abc", "abc", "abc"),
+          Seq(3, 6, 9).forall(input.drops.contains)
```
lihaoyi: For this, and subsequent tests, can we use a more exact `==` check instead of a `forall`/`contain` check? These examples are small enough we should be able to enforce *exactly* where the drops are happening.

fastparse/shared/src/test/scala/fastparse/iterator/IteratorTests.scala
```diff
@@ -0,0 +1,105 @@
...
+
+      'either {
+        val p = P( "a" ~ "b" ~ "c")
+        val either = P( (p ~ End) | ("abc" ~ p ~ End) | ("abcabc" ~ p ~ End))
+
+        val input1 = toInput("abcabcabc")
+        val Parsed.Success(_, i1) = either.parseInput(input1)
+        assert(
+          i1 == 9,
+          Seq(6, 7, 8, 9).forall(input1.drops.contains)
```
lihaoyi: It's not obvious to me why there should be drops at indices 6, 7, 8 and 9 in this test. Presumably it's because it's the "last" item in the `|` chain, but a comment would do great here to explain that fact

utils/shared/src/main/scala/fastparse/Utils.scala
```diff
@@ -1,12 +1,13 @@
 package fastparse
 
+import java.io.InputStream
+
 import scala.annotation.{switch, tailrec}
-import acyclic.file
+//import acyclic.file
```
lihaoyi: Looks like you had problems with the dependency-cycle-checker! Your problem is you are depending on `ParserInput` from `Utils.query`, but presumably also depending on `Utils` from `ParserInput`. The solution here is to make `ParserInput[T]` implement a smaller `IsReachable[T]` interface defined together with `Utils`, and make `query` only depend on the `IsReachable` interface

fastparse/shared/src/test/scala/fastparse/iterator/IteratorTests.scala
lihaoyi: In fact, if you could add a short explanation of *every* test case in this suite explaining what the expected behavior is and *why*, that would be great!

fastparse/shared/src/test/scala/fastparse/iterator/IteratorTests.scala
```diff
@@ -0,0 +1,105 @@
...
+        val pp = P( "a" ~ "b" ~ "c" ~ End)
+
+        val all = P( pp | (np ~/ np) | p ~ "e" | "abded".! )
+
+        val input = toInput("abded")
+
+        val Parsed.Success(res, _) = all.parseInput(input)
+        assert(res == "abded")
+      }
+    }
```
lihaoyi: Could we add one some small test cases to test the behavior of `dropBuffer` using the `WhitespaceApi`? Given that that was the cause of some bugs and frustration, we need to make sure that it keeps working.

As part of that, I would like a test case for the whitespace-ignoring `a ~ (b ~ (Pass ~/ Pass))` case. As discussed in chat, this currently would not correctly `dropBuffer` at the `~/`. Regardless we should still verify that fact, and if/when we decide to fix this behavior, we can update the test

perftests/shared/src/main/scala/perftests/Utils.scala
vovapolu: We can, but it could take a lot of time to do all of these benchmarks.

lihaoyi: Oh ok, let's leave them out then

fastparse/shared/src/test/scala/fastparse/iterator/IteratorTests.scala
vovapolu: Yeah, it's a good idea.

PR #99, opened 2016-08-07
Added a clarification to the docs about trailing whitespace

laughedelic:
See [discussion on gitter](https://gitter.im/lihaoyi/fastparse?at=57a54072857442dc0f575f68).

### DIFF
--- a/fastparse/shared/src/test/scala/fastparse/WhiteSpaceMathTests.scala
+++ b/fastparse/shared/src/test/scala/fastparse/WhiteSpaceMathTests.scala
@@ -7,12 +7,13 @@ import utest._
  * Same as MathTests, but demonstrating the use of whitespace
  */
 object WhiteSpaceMathTests extends TestSuite{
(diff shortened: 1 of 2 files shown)


fastparse/shared/src/test/scala/fastparse/WhiteSpaceMathTests.scala
```diff
@@ -7,19 +7,21 @@ import utest._
  * Same as MathTests, but demonstrating the use of whitespace
  */
 object WhiteSpaceMathTests extends TestSuite{
-  val White = WhitespaceApi.Wrapper{
+  val White = fastparse.WhitespaceApi.Wrapper{
```
laughedelic: This is not needed here to compile, but as it is copied to the docs without the package context, one cannot just copy this piece of code (it's not clear where does this `WhitespaceApi` come from).

He-Pin:
@laughedelic How about update this and get it merged:)

PR #100, opened 2016-08-07
Strange usage of Pass in scalaparse

laughedelic:
Hi, I asked about usage of `Pass ~` in scalaparse [on gitter](https://gitter.im/lihaoyi/fastparse?at=57a04f31c915a0e426b51aba) some days ago. So I was curious and tried to remove them. Here are those that don't fail tests. But there is still one left in `Types.scala` on the [line 33](https://github.com/laughedelic/fastparse/blob/bdade4a4c22de36264a9e076e402c517557a4087/scalaparse/shared/src/main/scala/scalaparse/Types.scala#L33) and one in `Scala.scala` on the [line 14](https://github.com/laughedelic/fastparse/blob/bdade4a4c22de36264a9e076e402c517557a4087/scalaparse/shared/src/main/scala/scalaparse/Scala.scala#L14). If I remove them some tests fail. 

I don't understand why they were there in the first place and why those two are essential. My guess would be that these two interact somehow with the cuts in `rep`, but it's just a speculation. I would appreciate if somebody who is more familiar with the internals of fastparse explained this.

### DIFF
--- a/scalaparse/shared/src/main/scala/scalaparse/Types.scala
+++ b/scalaparse/shared/src/main/scala/scalaparse/Types.scala
@@ -13,7 +13,7 @@ trait Types extends Core{
     P( (`private` | `protected`) ~ AccessQualifier.? )
   }
   val Dcl: P0 = {
-    P( Pass ~ ((`val` | `var`) ~/ ValVarDef | `def` ~/ FunDef | `type` ~/ TypeDef) )
+    P( (`val` | `var`) ~/ ValVarDef | `def` ~/ FunDef | `type` ~/ TypeDef )
   }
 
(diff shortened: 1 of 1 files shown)


lihaoyi:
Looks like most of those date back to https://github.com/laughedelic/fastparse/commit/8147158dacc02b4e01e64fbb243bfbf8fd8a29f1

From what I can tell, they were a vestigal thing to try and keep that diff "minimal": I wanted to get rid of the `WL` use sites but didn't want to change too much, so rather than deleting them entirely I replaced them with `Pass`.

Whatever it is, I'm confident enough in our test suite (positive and negative) that if everything passes let's kill 'em

PR #101, opened 2016-08-10
Added Parser.unapply(s: Repr): Option[T]

volth:
Why does ```Parser``` not have ```.unapply(s: Repr): Option[T]```?
It would allow to use it like regexp in pattern matching to get into ```String```|```Array[Byte]``` fields of complex objects, etc.
It looks trivial to add.


byteparse/shared/src/test/scala/byteparse/ByteTests.scala
```diff
@@ -116,6 +116,27 @@ object ByteTests extends TestSuite {
 
         val Parsed.Success(Some(Array(2)), 4) = captureOpt.parse(strToBytes("01 01 01 02"))
       }
+      'unapply {
+        val capture1 = P(BS(1).rep.! ~ BS(2) ~ End)
+
+        val capture1(Array(1, 1, 1)) = strToBytes("01 01 01 02")
```
vovapolu: Why do you use `Array` factory instead of `BS` or `ByteSeq` function? They are simple aliases of `Array[Byte]`, but their purpose is actually to replace explicit `Array` construction and improve readability (not only in `Parser` describing :) )

fastparse/shared/src/main/scala/fastparse/core/Parsing.scala
```diff
@@ -374,6 +374,26 @@ trait Parser[+T, ElemType, Repr] extends ParserResults[T, ElemType] with Precede
...
+   *  {{{
+   *  val p1 = CharIn('a'to'z').! ~ CharIn('0'to'9').rep(1).!.map(_.toInt)
+   *  List("a42x") match {
+   *    case p1(x: String, y: Int) :: _ => // x is "a", y is 42
+   *  }
+   * }}}
+   */
+  def unapply[Repr <% IndexedSeq[ElemType]](input: Repr): Option[T] = {
+    // there is no way to se an error message so `formatter` will not be used
+    parse(input)(null) match {
```
vovapolu: `null` is pretty dangerous here, I think it would be better to create some stub class instead. But anyway it won't work at all with last changes in my PR :D

fastparse/shared/src/main/scala/fastparse/core/Parsing.scala
```diff
@@ -374,6 +374,26 @@ trait Parser[+T, ElemType, Repr] extends ParserResults[T, ElemType] with Precede
...
+   *  For example:
+   *
+   *  {{{
+   *  val p1 = CharIn('a'to'z').! ~ CharIn('0'to'9').rep(1).!.map(_.toInt)
+   *  List("a42x") match {
+   *    case p1(x: String, y: Int) :: _ => // x is "a", y is 42
+   *  }
+   * }}}
+   */
+  def unapply[Repr <% IndexedSeq[ElemType]](input: Repr): Option[T] = {
```
vovapolu: I've just compiled code like this: 
```scala
 def unapply(input: Repr)(implicit formatter: ElemTypeFormatter[ElemType],
                                    converter: ResultConverter[ElemType, Repr]): Option[T] = {
    parse(input) match {
      case Parsed.Success(r, _) => Some(r)
      case _ => None
    }
```
And it seems everything is okay, could you explain the problem with implicits in unapply more detailed?

byteparse/shared/src/test/scala/byteparse/ByteTests.scala
volth: This test block is just a slightly modified copy-paste of the block few lines above.

fastparse/shared/src/main/scala/fastparse/core/Parsing.scala
volth: You are right, the implicit could be back.

byteparse/shared/src/test/scala/byteparse/ByteTests.scala
vovapolu: Oh :) Okay, I will try to change them to `BS`.

fastparse/shared/src/main/scala/fastparse/core/Parsing.scala
vovapolu: Actually if you look at my new [PR](https://github.com/lihaoyi/fastparse/pull/97), `parse` method definition has been changed a bit. Could you wait until it's moved to the master and then make your `unapply` method satisfy new changes? Code that I wrote above is a good example of how to modify it.

volth: Sure. Having ```parse(Repr)``` instead of ```parse(IndexedSeq[ElemType]``` would make it even more simpler, with no ```<%```.

I am afraid it would confusing to use ```unapply``` with ```Iterator``` because one may write
```
input match {
  case parser1(x) =>
  case parser2(x) =>
}
```
reusing same input for another ```parse```.
Perhaps ```unapply``` could accept ```Iterable``` and create ```Iterator``` for each parse.

vovapolu: Or maybe just don't accept `Iterator` at all ;) Parsing of `Iterator` is unstable feature right now, and I can't imagine use case where we want to pattern-match `Iterator` or even relatively big input (1Kb or more).

volth: I do not have a real use case either. I was thinking about the inputs which are ```Iterable[Byte]``` or ```Iterable[Char]``` but not ```String``` or ```Array[Byte]``` (```Vector[Byte]```, etc).

lihaoyi:
Thanks for the diff @volth, and for helping review @vovapolu!

In general, I think this is a great idea. I won't really look closely until #97 lands, since that shakes up a lot of things and is a big diff in it's own right, but after that goes in I'll give this another look and work on merging it

fastparse/shared/src/main/scala/fastparse/core/Parsing.scala
vovapolu: You can add any `Repr` type as you want by simply implementing new instance of `ResultConverter` and the case when we don't know the actual type of input data or these types are varying from run to run is not so big loss.

volth: I'd say, with the library providing ```parse(input: Iterator[Byte])``` people rather write ```parse(mybytes.toIterator)``` then would deal with implementing implicit converters.

vovapolu:
@volth Hi! The PR was merged, so you can refresh your code and I will merge it.

vovapolu:
@volth Great idea and work :)

PR #104, opened 2016-08-21
Improve aggressiveness of dropBuffer calls caused by cuts

lihaoyi:
In principle, we should be able to call `dropBuffer` immediately after the `lhs` of `lhs ~/ rhs` succeeds, as the cut occurs "in between" the `lhs` and `rhs`. 

Currently, we only call `dropBuffer` after the `rhs` succeeds, as when we're evaluating the "current" `lhs ~/ rhs`, after `lhs` succeeds we decide whether or not to `dropBuffer` based on the *previous* existence of cut, ignoring the fact that a cut had just occurred. Thus, if we had `a ~/ b ~ c`, we would call `dropBuffer` after `b` completes, while we really want it to occur after `a`.

This diff makes the consideration of `dropBuffer` also consider whether the "current" `CustomSequence`/iteration-in-`Sequence.Flat` just performed a cut. Thus, given `a ~/ b ~ c`, we now call `dropBuffer` immediately after `a`, in addition to after `b`.

Also does some miscellaneous changes:

- Convert `LoggedDropsParserInput` to use `SortedSet` and give it a nice
  `toString`, so it's easier to debug things when they misbehave

- Add a unit test to verify that top-level `~`s call `dropBuffer`, even
  without any cuts. This behaves as expected right now, but good to have
  a dedicated test for it

- Add unit tests to verify the new behavior

Tests seem to pass https://travis-ci.org/lihaoyi/fastparse/builds/153880403, at least as-passing-as-they-currently-are

Review by @vovapolu

### DIFF
--- a/fastparse/shared/src/test/scala/fastparse/iterator/IteratorTests.scala
+++ b/fastparse/shared/src/test/scala/fastparse/iterator/IteratorTests.scala
@@ -11,17 +11,40 @@ object IteratorTests extends TestSuite {
                                         (implicit formatter: ElemTypeFormatter[ElemType])
     extends IteratorParserInput[ElemType](data) {
 
-    var drops = mutable.Set.empty[Int]
+    val drops = mutable.SortedSet.empty[Int]
 
     override def dropBuffer(index: Int): Unit = {
       drops.add(index)
       super.dropBuffer(index)
     }
+    override def toString = s"LoggedDropsParserInput($drops)"
   }
 
   def toInput(string: String) = new LoggedDropsParserInput[Char](string.grouped(1).map(wrapString))
 
   val tests = TestSuite {
+
+    'immediateCutDrop{
+      import fastparse.all._
+      val p = P( "ab" ~/ "cd" | "z" )
+
+      val input = toInput("abcdef")
+      val Parsed.Success(res, i) = p.parseInput(input)
+      // Make sure that we drop immediately at position 2, since that is where
+      // the cut has taken place, rather than at position 4 as we did earlier.
+      assert(input.drops == Set(2, 4))
+    }
+    'topLevelNoCuts{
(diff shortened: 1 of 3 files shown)


fastparse/shared/src/test/scala/fastparse/iterator/IteratorTests.scala
```diff
@@ -11,17 +11,55 @@ object IteratorTests extends TestSuite {
...
+      val Parsed.Success(res, i) = p.parseInput(input)
+      // Make sure that we drop immediately at position 2, since that is where
+      // the cut has taken place, rather than at position 4 as we did earlier.
+      assert(input.drops == Set(2, 4))
+    }
+    'whitespaceImmediateCutDrop{
+      import fastparse.noApi._
+      val White = fastparse.WhitespaceApi.Wrapper{
+        import fastparse.all._
+        NoTrace(" ".? ~ " ".rep) // note that the whitespace delimiter has cut
```
vovapolu: Hmm, whitespace delimiter hasn't cut in this case

vovapolu:
Yes, that's true. I've missed it when I was thinking about placement of `dropBuffers`.

vovapolu:
@lihaoyi Great :) Let's merge and look how your unit-test PR is going to work with new `dropBuffer` improvements.

lihaoyi:
FTR, this reduced the Max Buffer Size of the iterator parsers in our various benchmarks by small amounts:

Benchmark  | Before | After
-------------|---------|-----
ScalaParse  | 1587 | 1555
PythonParse | 2025 | 2006
ByteParse | 40 | 36

CSSParse wasn't affected; I guess we need to put some cuts in that.

These numbers are all for batch-size 1, i.e. the "exact" buffer size required, without any real buffering.

Relatively small improvement overall, but non-zero

PR #105, opened 2016-08-21
Add a *lot* more tests for our iterator parser

lihaoyi:
Since basically any parser is expected to work with iterator input, this diff embeds iterator tests into the success/fail test cases of:

- `ParsingTests`
- `MathTests`
- `JsonTests`
- `scalaparse`
- `pythonparse`
- `BmpTests`
- `ClassTests`

The tests are sometimes not quite identical, e.g. for failure tests `parseIterator` does not support `.traced.trace` since it would require a second parse on the already-exhausted iterator, but they should provide enough checks (e.g. error index) that we can be more confident it works and keeps working. We run the tests with iterators of various chunks-sizes to try and find any bugs (hopefully none).

I also deleted `JsonIteratorTests`, as it's obsolete now that `JsonTests` tests iterator behavior, and the `summingbird` scalaparse test since that repo no longer exists on github

Some of the tests passed locally, let's see if any cause problems in CI. This should in theory make a bunch of tests slower, hopefully the end result won't be *too* slow...

Review by @vovapolu.


byteparse/shared/src/test/scala/byteparse/BmpTests.scala
```diff
@@ -67,31 +68,30 @@ object BmpTests extends TestSuite {
...
+              Pixel(BS(0, 0, 0xff, 0x7f)),
+              Pixel(BS(0, 0, 0xff, 0x7f))
             )
           )
         )
+
+        val Parsed.Success(bmp2, _) = bmp.parse(file1)
+        assert(compareBmps(bmp2, expected))
+
+        for(chunkSize <- Seq(1, 4, 16, 64, 256, 1024)){
```
vovapolu: Iterator parsing is approximately 2x slower than non-Iterator, so adding 6 more Iterator tests may significantly increase total time of all of these tests.

lihaoyi: Yeah let's see how long it takes in travis. If it ends up taking too long, I'll disable these iterator tests for all the `ProjectTests` and leave them running for unit-test-style tests. Those run basically instantaneously and making them 13x slower should be fine

vovapolu:
@lihaoyi Now it has conflicts with master, probably because of the merging with the last PR.

lihaoyi:
I merged it on top of the last PR, let's see what travis says.

When we do end up merging this, let's use the "Merge and Squash" button, since this chain of commits is getting a bit long and messy

lihaoyi:
I went and killed a bunch of the unnecessary intermediate builds on travis-ci, but even so this build probably won't be done for another 2 hours or so given how slow things are (I left the merge-to-master test run to run before this) so I'll probably come check back in a few hours/tomorrow to see what happened

lihaoyi:
@vovapolu the last build https://travis-ci.org/lihaoyi/fastparse/builds/153907677 seems like it is green-ish enough

PR #110, opened 2016-08-22
Shard the things we're running onto different travis nodes

lihaoyi:
So they can work-steal from each other (e.g. a fast-finishing node can grab more work to do) and we don't need to dig through so much spam to find what we want (e.g. when something fails)

This does result in a lot of duplication in terms of compiling and re-compiling `fastparseJS` and `fastparseJVM`. It's unclear how much of a slowdown this will be. 

I kicked off a build https://travis-ci.org/lihaoyi/fastparse/builds/154041707. The status quo is that things take about 1 hour to 2 hours, with a few builds timing out at 2 hours. Let's see how this compares

The commit history is really gross, please squash on merge!

EDIT: also reformatted a bunch of the stuff in `build.sbt` that shouldn't change anything, apart from adding `parallelExecution := false` to our perf tests


build.sbt
```diff
@@ -132,36 +143,46 @@ lazy val cssparse = crossProject
...
 
-lazy val perftests = crossProject.settings(
-   name := "perfomance-tests"
- ).settings(shared:_*).dependsOn(fastparse, pythonparse, scalaparse, cssparse, byteparse)
+lazy val perftests = crossProject
+  .dependsOn(fastparse, pythonparse, scalaparse, cssparse, byteparse)
+  .settings(shared:_*)
+  .settings(
+    name := "perfomance-tests",
+    parallelExecution := false
```
lihaoyi: This bit should let us run the perftests all in the same process without worrying about parallelism messing with our resutls

lihaoyi:
Looks like we're down from 105 minutes on master to 33 minutes on the last build on this branch (https://travis-ci.org/lihaoyi/fastparse/builds/154070004), hopefully without too much loss of coverage.

Review by @vovapolu

vovapolu:
@lihaoyi Well, I don't fully understand these things about Travis and build configs, but it seems that everything's okay :) Now at least it doesn't get timeout in Futures.

lihaoyi:
> Well, I don't fully understand these things about Travis and build configs

Then it's a great opportunity to learn =D Even if you don't understand now, take the chance to read up until you have some idea of what the various things I'm changing are, even if it's not a complete understanding.

vovapolu:
I mean I am not a good person for code review ;) But I will read it, yeah.

PR #113, opened 2016-08-25
Updating the docs

vovapolu:
With recent global changes in fastparse core and adding big parsers (cssparse, byteparse) the docs should be updated with guides for writing byte parsers and descriptions and interactive examples of new parsers.


readme/Readme.scalatex
```diff
@@ -322,6 +335,89 @@
...
+                        @p
+                            @hl.scala{StringIn} becomes @hl.scala{SeqIn}
+                        @i
+                            It isn't a simple replacement of substrings, and the reason for this is described in @sect.ref{Inner Structure}.
+
+             @p
+                For instance, the basic parser with @hl.scala{AnyByte}
+             @hl.ref(bytetests/"ByteTests.scala", Seq("'anychar", ""))
+
+    @sect{Inner Structure}
```
lihaoyi: Can we move this section under `Internals`?

readme/Readme.scalatex
```diff
@@ -322,6 +335,89 @@
...
                 @hl.ref(tests/"ExampleTests.scala", Seq("'composenocut", ""))
+        @sect{Unapply}
+            @p
+                @hl.scala{Parser} class has (as many other classes in Scala) a convenient method @hl.scala{unapply}
+                which allows to do pattern matching using parsers (like regular expressions) in your code.
+
+                @hl.ref(tests/"ExampleTests.scala", "val capture1", "val captureOpt")
+
+
+    @sect{Writing Byte Parsers}
```
lihaoyi: Let's move this, together with the example byte parsers, to a new top-level section next to `Fastparse` and `ScalaParse` and `Changelog`. For now, the people using byte-parsers will probably be pretty different from the people using string-parsers, so giving them a separate top-level section probably makes sense.

readme/Readme.scalatex
```diff
@@ -463,31 +559,147 @@
...
+                It's able to retrieve almost full information about given class including all methods, fields, subclasses, code
+                and pack it into the convenient AST.
+                On the similarity with other big parsers it has good set of unit-tests
+                with tests which compile real projects from github (joda-time, junit4, jenkins and etc.)
+                and check each built file.
+
     @sect{API Highlights}
 
-        @sect{Parser[T]}
+        @sect{Parser[T, ElemType, Repr]}
```
lihaoyi: No need to change this for now; the `Parser` type from `fastparse.all` is still `fastparse.all.Parser[T]`. We can talk about the more general `Parser[T, ElemType, Repl]` under the `Internal Structure` section

readme/Readme.scalatex
```diff
@@ -782,6 +1000,10 @@
             "com.lihaoyi" %%% "scalaparse" % "@fastparse.Constants.version"
 
 @sect{Change Log}
+    @sect{0.4}
+        @ul
+            @li
+                A lot of things :)
```
lihaoyi: Let's take the effort to flesh this out =) This will basically be the summary of your GSOC work, so you should let everyone know what you did!

readme/Readme.scalatex
```diff
@@ -463,31 +559,147 @@
...
+                    and the easiest and fastest method to parse them is to write @hl.scala{.rep(exactly=...)}.
+                @li
+                    @b
+                        The same extensive usage of @hl.scala{.flatMap}.@br
+                    @hl.scala{flatMap} allows to retrieve data from one parser and pass it to the next.
+                    The most popular and primitive example is dynamic-sized array with length written at the beginning.
+                    Similar algorithm is used in the @hl.scala{bmp} when header information is passed to the @hl.scala{bmpRow}
+                    that returns new parser for row in this particular bmp file.
+
+        @sect{ClassParser}
```
lihaoyi: As mentioned in my comment on `Writing Byte Parsers`, let's move this to the new top-level byte-parser section

readme/Readme.scalatex
```diff
@@ -322,6 +335,89 @@
...
+                    @hl.scala{ElemPred} for @hl.scala{CharPred} and @hl.scala{BytePred}
+                @li
+                    @hl.scala{ElemIn} for @hl.scala{CharIn} and @hl.scala{ByteIn}
+                @li
+                    @hl.scala{ElemsWhile} for @hl.scala{CharsWhile} and @hl.scala{BytesWhile}
+                @li
+                    @hl.scala{SeqIn} for @hl.scala{StringIn}
+            but you can, of course, use these abstract versions of names wherever you want.
+
+    @sect{Iterator Parsers}
```
lihaoyi: Let's call this `Iterator Parsing` instead of `Iterator Parsers`, because this applies to all existing and future parsers, not just those specifically written to work with iterators.

readme/Readme.scalatex
```diff
@@ -322,6 +335,89 @@
...
+        @p
+            In addition to the regular plain parsing you can simply parse streaming data as @hl.scala{Iterator}.
+            Just call @hl.scala{.parseIterator} instead of @hl.scala{.parse} in your parser and pass the @hl.scala{Iterator}.
+            It will try to optimize the memory consumption using the information
+            about cuts, capturing and other special behaviors of simple parsers,
```
lihaoyi: Let's link this to the relevant docs via `@sect.ref{Cuts}`

readme/Readme.scalatex
```diff
@@ -322,6 +335,89 @@
...
+            but it's quite a complicated logic described in source docs.
+            Anyway, you can use the simple rule to improve the performance in the Iterator parsing -
+            every cut in your parser possibly reduces the total memory consumption,
+            so if you want to write good Iterator parser just add cuts wherever it's possible.
```
lihaoyi: This section could use some concrete numbers: 

- How much does iterator parsing affect the size of the buffer? We should have a table showing the maximum buffer sizes of {ScalaParse, PythonParse, JsonParse} when using `.parseIterator`, compared against the size of the "full" file that we would have loaded using `parse(s: String)`.

- How is performance? Compared to String parsing, and comparing various sizes of buffer for some example parser (I'd pick Scalaparse as an example), how much is the slowdown?

Furthermore, let's mention somewhere that `@sect.ref{Tracing}` is not available while using iterator parsers, as the iterator gets exhausted and cannot be used a second time

readme/Readme.scalatex
```diff
@@ -322,6 +335,89 @@
...
+                    @hl.scala{SeqIn} for @hl.scala{StringIn}
+            but you can, of course, use these abstract versions of names wherever you want.
+
+    @sect{Iterator Parsers}
+        @p
+            In addition to the regular plain parsing you can simply parse streaming data as @hl.scala{Iterator}.
+            Just call @hl.scala{.parseIterator} instead of @hl.scala{.parse} in your parser and pass the @hl.scala{Iterator}.
+            It will try to optimize the memory consumption using the information
+            about cuts, capturing and other special behaviors of simple parsers,
+            but it's quite a complicated logic described in source docs.
```
lihaoyi: Let's not say `it's quite complicated logic`: instead, let's try to explain what it does.

Parsing iterators, it maintains an internal buffer of the iterator segments it has read so far. It then uses Cuts to determine when the parser no longer can backtrack past certain points, and discards the parts of buffered input it no longer needs. This discarding happens after approximately after every `~` or iteration of `.rep`.

readme/Readme.scalatex
```diff
@@ -322,6 +335,89 @@
...
+                    @hl.scala{ElemIn} for @hl.scala{CharIn} and @hl.scala{ByteIn}
+                @li
+                    @hl.scala{ElemsWhile} for @hl.scala{CharsWhile} and @hl.scala{BytesWhile}
+                @li
+                    @hl.scala{SeqIn} for @hl.scala{StringIn}
+            but you can, of course, use these abstract versions of names wherever you want.
+
+    @sect{Iterator Parsers}
+        @p
+            In addition to the regular plain parsing you can simply parse streaming data as @hl.scala{Iterator}.
```
lihaoyi: We should mention somewhere that we use chunked `Iterator[String]` or `Iterator[Array[Byte]]` instead of `Iterator[Char]`/`Iterator[Byte]` for performance, and compatibility with most streaming input sources (reading from file, network, etc.) which also chunk their input. The size of the chunks affects the performance/max-memory-usage of the parser, which they can see for themselves in the table below

readme/Readme.scalatex
```diff
@@ -322,6 +335,89 @@
...
+            The first is new word @hl.scala{BS} or @hl.scala{ByteSeq} that merely means alias for @hl.scala{Array[Byte]},
+            the second is new function for convenient conversion from @hl.scala{String} to @hl.scala{Array[Byte]} - @hl.scala{strToBytes}.
+        @p
+            Of course, you aren't restricted to use only them, you are free to create Array of Byte by any method or function,
+            but, naturally, it's recommended to follow the one uniform style and @hl.scala{BS} with @hl.scala{strToBytes} is the good example of it.
+
+        @sect{New Parsers}
+            @p
+                Byte Parsers doesn't bring anything new to the basic set of parsers.
+                @hl.scala{Option}, @hl.scala{Repeat}, @hl.scala{Capture} and ect. are still here with the same logic,
```
lihaoyi: These should all be `@sect.ref`s to link to the relevant sections of the string-parser docs

readme/Readme.scalatex
```diff
@@ -322,6 +335,89 @@
...
+
+                @hl.ref(tests/"ExampleTests.scala", "val capture1", "val captureOpt")
+
+
+    @sect{Writing Byte Parsers}
+        @p
+            In the same way as the @hl.scala{String} parsers we can build @hl.scala{Array[Byte]} or simply @hl.scala{Byte} parsers.
+        @p
+            The first example will be also about parsing a single byte sequence
+        @hl.ref(bytetests/"ByteTests.scala", Seq("'sequence", ""))
```
lihaoyi: Could we make this example pass in a raw `Array[Byte]` instead of using `strToByte`?

readme/Readme.scalatex
```diff
@@ -463,31 +559,147 @@
...
+                If you haven't any bmp around, you can download classic
+                @a("lena", href:="https://raw.githubusercontent.com/lihaoyi/fastparse/master/byteparse/jvm/src/test/resources/lena.bmp").
+
+            @p
+                @hl.scala{BmpParser} is a good first example of byte parser, it's not small, but structure is pretty simple.
+
+            @i
+                @hl.scala{AnyWordI} and @hl.scala{AnyDwordI} are just parsers that consumes any word (2 bytes) or dword (4 bytes).
+
+            @hl.ref(bytemain/"BmpParser.scala", "val fileHeader")
```
lihaoyi: Is `AnyWordI` and `AnyDWordI` part of the public API? If not, we should make them part of the public API and describe them under `Writing Byte Parsers`. An alternative is to define them directly in `BmpParser.scala`, or rip them out of `BmpParser.scala`  and replace them with the respective `.!` parsers

readme/Readme.scalatex
```diff
@@ -463,31 +559,147 @@
...
+
+            @p
```
lihaoyi: These few paragraphs would be much better if placed next to the code they're describing.

Instead of having separate paragraphs, let's include them in the code snippet as comments.

The order of defining the parser does not matter, so you can re-order it to e.g. move `val bmp` to the top of the example, in order to maintain the same order of your explanation

readme/Readme.scalatex
vovapolu: Hm, it's a piece of code down there, that is `Parser[T, ElemType, Repr]` class definition.

readme/Readme.scalatex
vovapolu: I think two new concepts is not a problem for understanding :) Let's slightly rewrite it with adding comments for explanation of them.

lihaoyi:
Looks good to me, thanks!

PR #115, opened 2016-08-27
Speed up Iterator parsers by swapping out Vector for Circular Buffer

lihaoyi:
Looking up arrays are faster than vectors, copying arrays is faster than copying vectors, mutating arrays is faster than copy-and-updating vectors, and arrays take less memory to store than vectors

JProfiler was showing about 20-30% of time being spent inside various vector operations (lookup, copy-and-update, copy-to-array, ...) and this should make it disappear.

This implements a growable circular buffer to use as a queue, to replace the old var/Vector. As we write stuff to the buffer, we shift the `end` index to the right, wrapping around if there's empty space at the start of the `data` array. As we drop stuff, we simply shift the "start" index from to the right as well, also wrapping around. If the end index ever catches up with the start index, it means the buffer is full and we copy everything to a large array.

This means that the buffer only grows as necessary, and only allocates if it grows. If we are writing/dropping constantly, the two start and end indices will go around and around the data array, but the data array itself will never need to be re-allocated. The buffer only grows and never shrinks, which should be fine for now since it'll never grow much bigger than the input size, and will get discarded once the parse run is complete

Roughly copied from https://github.com/lihaoyi/SprayWebSockets/blob/master/src/main/scala/spray/can/server/websockets/UberBuffer.scala and modified to fit (needs to work on both char/byte, no more input/outputstream API, ...)

This doesn't seem to have much effect on small 1-2 length chunks (maybe even a small slowdown?) but then it seems to have a pretty consistent ~30% speedup on larger chunk sizes. The vector-related bottlenecks also disappeared from the JProfiler profile

# Benchmarks:

I just ran the benchmarks using BmpParse, but presumably it would affect other parsers as well

## Master
| Runs\Chunk | 1           | 2           | 4           | 16          | 64          | 1024        | 4096        | 
|------------|-------------|-------------|-------------|-------------|-------------|-------------|-------------| 
| 1          | 20          | 31          | 35          | 51          | 55          | 60          | 60          | 
| 2          | 27          | 35          | 35          | 55          | 60          | 58          | 59          | 
| 3          | 28          | 33          | 29          | 50          | 56          | 58          | 52          | 
| 4          | 25          | 33          | 33          | 50          | 57          | 55          | 58          | 
| 5          | 27          | 38          | 35          | 56          | 43          | 51          | 55          | 
| Average    | 25.4        | 34          | 33.4        | 52.4        | 54.2        | 56.4        | 56.8        | 
| Std Dev    | 3.20 | 2.64 | 2.60 | 2.88 | 6.53 | 3.50 | 3.27 | 

## Uber Buffer
| Runs\Chunk | 1           | 2           | 4           | 16          | 64          | 1024        | 4096        | 
|------------|-------------|-------------|-------------|-------------|-------------|-------------|-------------| 
| 1          | 20          | 33          | 48          | 76          | 79          | 74          | 79          | 
| 2          | 21          | 33          | 46          | 71          | 80          | 83          | 84          | 
| 3          | 23          | 35          | 45          | 69          | 75          | 77          | 66          | 
| 4          | 17          | 35          | 42          | 68          | 74          | 76          | 72          | 
| 5          | 17          | 31          | 36          | 59          | 67          | 67          | 74          | 
| Average    | 19.6        | 33.4        | 43.4        | 68.6        | 75          | 75.4        | 75          | 
| Std Dev    | 2.60 | 1.67 | 4.66 | 6.18 | 5.14  | 5.77 | 6.85   | 

## Comparison

| Chunk      | 1           | 2           | 4           | 16          | 64          | 1024        | 4096        | 
|------------|-------------|-------------|-------------|-------------|-------------|-------------|-------------| 
| Speedup    | 0.77 | 0.98 | 1.29 | 1.30 | 1.38 | 1.33 | 1.32 |

### DIFF
--- a/utils/shared/src/main/scala/fastparse/UberBuffer.scala
+++ b/utils/shared/src/main/scala/fastparse/UberBuffer.scala
@@ -1,0 +1,103 @@
+package fastparse
+
+import scala.reflect.ClassTag
+
+/**
+  * A very fast circular, growable read-write byte buffer.
+  */
+class UberBuffer[T: ClassTag](initSize: Int = 32){ self =>
+  private[this] var data = new Array[T](initSize)
+  private[this] var readPos = 0
+  private[this] var writePos = 0
+
+  def capacity = data.length
+  def apply(index: Int) = {
+    if (index < 0 || index >= length) throw new IndexOutOfBoundsException(
+      s"UberBuffer index $index must be between 0 and $length"
+    )
+    data((index + readPos) % capacity)
+  }
+  def length = {
+    if (writePos == readPos){
+      0
+    }else if(writePos > readPos){
+      //   1 2 3
+      // - - - - -
+      //   r   W
+      //       R
+      writePos - readPos
+    } else {
+      // 3 4   1 2
+      // - - - - -
+      //   W   r
+      //   Rs
+      data.length - readPos + writePos
+    }
+  }
+  private[this] def writeAvailable = {
+    if (writePos == readPos){
+      data.length - 1
+    }else if (writePos > readPos){
+      //    1 2 3 4
+      //  - - - - -
+      //  W R w
+      data.length - writePos - 1 + readPos
+    }else{
+      //    1
+      //  - - - - -
+      //    w W R
+      readPos - writePos - 1
+    }
+  }
+
+  private[this] def expand() = {
+    val newData= new Array[T](data.length * 2)
+
+    if (readPos <= writePos){
+      System.arraycopy(data, readPos, newData, 0, writePos - readPos)
+      writePos = writePos - readPos
+    }else{
+      System.arraycopy(data, readPos, newData, 0, data.length - readPos)
+      System.arraycopy(data, 0, newData, data.length - readPos, writePos)
+      writePos = writePos + data.length - readPos
+    }
+    readPos = 0
+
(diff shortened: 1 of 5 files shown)


utils/shared/src/main/scala/fastparse/UberBuffer.scala
```diff
@@ -0,0 +1,103 @@
+package fastparse
+
+import scala.reflect.ClassTag
+
+/**
+  * A very fast circular, growable read-write byte buffer.
+  */
+class UberBuffer[T: ClassTag](initSize: Int = 32){ self =>
```
vovapolu: Why is `ClassTag` necessary here?

vovapolu:
Do we want to update a readme benchmarks after this PR?

utils/shared/src/main/scala/fastparse/UberBuffer.scala
lihaoyi: `ClassTag` is necessary to instantiate a generic `new Array[T]` http://stackoverflow.com/a/16921379/871202

lihaoyi:
Sure, I'll update the benchmarks, but let's do so as a separate PR. I'm continuing to poke around and tweak things and that will probably have other perf impact in future

PR #117, opened 2016-08-29
Replace the tree-shaped Trie with a flat array-backed Trie with exactly the same behavior.

lihaoyi:
This was an attempt to speed things up by swapping out the Trie for another implementation. In practice, it didn't seem to make anything faster, but leaving this here because it's interesting and maybe someone could investigate why later

### DIFF
--- a/fastparse/shared/src/main/scala/fastparse/Api.scala
+++ b/fastparse/shared/src/main/scala/fastparse/Api.scala
@@ -6,6 +6,8 @@ import language.experimental.macros
 import fastparse.parsers.{Intrinsics, Terminals}
 import fastparse.Utils.HexUtils
 
+import scala.reflect.ClassTag
+
 /**
  * This is basically a trait which contains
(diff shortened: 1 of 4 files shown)


vovapolu:
Hmm, why it would work faster than the trie implementation? It also has random-access operations and it seems like you wrote the same tree but inside array :)

lihaoyi:
My theory was that having everything packed inside an array would improve the cache locality of the Trie: rather than following pointers all over main memory to look up each `TrieNode` object, the "pointers" would all be in a single contiguous Array, hopefully being closer together in memory, being paged in from main memory together, and being cached together inside the CPU so lookups are fast.

Overall, though, the profiles showed the TrieNode lookups being only about 2% of the overall runtime. That probably explains why there wasn't any real difference. I haven't actually benchmarked the two tries stand-alone to see if there's any difference in the trie-specific behavior, but if it doesn't affect the overall parser runtime I don't care that much

PR #118, opened 2016-08-30
Added fold method to Parsed ADT

### LINKED ISSUE #34: Pattern-matching on Result produces a (spurious?) warning

mgzuber:
This should mostly resolve #34 as users would not have to pattern match on a the Parsed result type

### DIFF
--- a/fastparse/shared/src/main/scala/fastparse/core/Parsing.scala
+++ b/fastparse/shared/src/main/scala/fastparse/core/Parsing.scala
@@ -32,6 +32,14 @@ sealed trait Parsed[+T, ElemType]{
     case s: Parsed.Success[T, ElemType] => s
     case f: Parsed.Failure[ElemType] => throw new ParseError[ElemType](f)
   }
+
+  /**
+    * Returns the result of $onSuccess if the parsing is successful, otherwise it returns the result of $onFailure
+    */
+  def fold[X](onFailure: (Parser[_, ElemType, _], Int, Parsed.Failure.Extra[ElemType]) => X, onSuccess: (T, Int) => X): X = this match {
+    case Parsed.Success(t, i) => onSuccess(t, i)
+    case f: Parsed.Failure[ElemType] => onFailure(f.lastParser, f.index, f.extra)
+  }
 }
 
 case class ParseError[ElemType](failure: Parsed.Failure[ElemType]) extends Exception(
--- a/fastparse/shared/src/test/scala/fastparse/ExampleTests.scala
+++ b/fastparse/shared/src/test/scala/fastparse/ExampleTests.scala
@@ -454,6 +454,21 @@ object ExampleTests extends TestSuite{
         assert(capturedString == expectedString)
       }
     }
+    'folding{
+      sealed trait AndOr
+      case object And extends AndOr
+      case object Or extends AndOr
+      val and = P(IgnoreCase("And")).map(_ => And)
+      val or = P(IgnoreCase("Or")).map(_ => Or)
+      val andOr = P(and | or)
+
+      def check(input: String, expectedOutput: String) =
+        assert(andOr.parse(input).fold((_, _, _) => s"Cannot parse $input as an AndOr", (v, _) => s"Parsed: $v") == expectedOutput)
+
+      check("AnD", "Parsed: And")
+      check("oR", "Parsed: Or")
+      check("IllegalBooleanOperation", "Cannot parse IllegalBooleanOperation as an AndOr")
+    }
 
   }
 }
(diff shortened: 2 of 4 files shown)


vovapolu:
Could you add one or two tests to show the example of usage for your new method?

mgzuber:
Absolutely. It was pointed out to me as well that method signature isn't the best either so I'm rethinking that a little too.

mgzuber:
@vovapolu I've added some tests to show some example usage, I can add more if need be. I changed the fold signature to use the type signatures of `Parsed.Success.unapply` and `Parsed.Failure.unapply` and additionally added another fold method whose `onFailure` input only takes the error message from parsing.

In my experience, when pattern matching a parsing result, in the case of a failure I only want the error message, so that is why I've included the second fold signature.

vovapolu:
@mgzuber Hmm, It looks quite complicated. User who just want transform result to some other value depending on whether parsing was successfully or not should now know about all of these stuff in `Failure` class. Mostly the pattern matching of `Parsed` looks like this
```scala
result match {
  case s: Parsed.Success => // doing something with s
  case f: Parsed.Failure => // doing nothing or reporting about an error
}
```
The inner structure of `Failure` is usually used only in the core and isn't needed for regular user. 

The initial idea is good, but I'm not sure that we need such non-trivial method in the basic class of fastparse. Probably we should think how to make it more clear for users who don't want to know all the subtle aspects of this library.

lihaoyi:
I don't really understand the signature of `onFailure`, or why there are two different versions of `fold`. Given that Failure is defined as `Failure(lastParser, index, extra)`, wouldn't it make sense to have fold take a `(lastParser, index, extra) => T` as the `onFailure` callback? I thought that's how fold normally worked: you take the things that are part of the case class and plomp them into the callback arguments 

If they want the `.msg` error message, they can construct it themselves with `index` and `lastParser`, or we could provide a helper on `extra` as `extra.msg` that does it for them. If they want the full traced/whatever, they have all that on `extra`. That should be enough for anything someone would want to do, since that's all `Failure` has on it in the first place

mgzuber:
I agree that the type signature of the `onFailure` method isn't great and sort of requires the user to deal with the internal details of the `Failure` class which isn't ideal (and also which is why I added a second fold method that only takes the error message). I do, however, like @lihaoyi suggestion of adding a helper method on `extra` and simplifying the signature of `onFailure` to use only the types in the case class. I also think that a signature for `onFailure` of `Parsed.Failure => X` could also be a good idea.

lihaoyi:
While I think having the two callbacks as `Failure => T` and `Success => T` might be convenient, that would be sufficiently different from other definitions of `fold` that I'd be reluctant to call it that. After all, `fold` isn't just a method we're defining ad-hoc, but a "standard" ADT-extracting thing that's defined on lists, options, and all the shapeless/scalaz stuff. In those cases, the callbacks always deal with the "extracted" contents of the thing being folded.

Thus I think we should extract the contents of the failure on fold and pass them to the callback, and same with the success callback. In the success/failure cases it's only 2/3 fields, so it's not like it's a lot of implementation details we're leaking

mgzuber:
I've changed the type signatures as per the comment above. I can add a function that uses the callbacks `Failure => T` and `Success => T` but I agree I'm not entirely sure what it should be called. I do think that `fold` could be appropriate because it would resemble the `fold` method defined on `Either` which resembles `Parsed` somewhat.

lihaoyi:
@mgzuber looks good

Last change I'd request is, as @vovapolu mentioned, an "example" unit test to show a nice stand-alone example of what `fold` does, without all the associated test infrastructure and helpers. This should live in https://github.com/lihaoyi/fastparse/blob/master/fastparse/shared/src/test/scala/fastparse/ExampleTests.scala, and linked into the docs probably at the bottom of http://www.lihaoyi.com/fastparse/#ParsingResults with a bit of explanation

mgzuber:
Done

vovapolu:
@mgzuber There's a conflict with your PR, probably it's because of recent readme splitting. Check the last commit in the master.

mgzuber:
I'll rebase on top of the most recent code

mgzuber:
Ok, I've resolved the conflicts

lihaoyi:
Looks good to me, thanks @mgzuber!

PR #119, opened 2016-08-31
Another pass of code/docs polish work for byte-parser API

lihaoyi:
- Rename `strToBytes` -> `hexBytes`, to make it clear what kind of st…ring it is

- `BytesParser` -> `HexBytesParser`

- Added forwarder values for `LE` and `BE` inside `fastparse.bytes` so you don't need to import them separately

- Converted a few `.flatMap`s to use `for`-comprehensions since I think those are nicer to read. Added some docs to mention that you can use both `for`-comprehensions and explicit `flatMap`s

- Added a new `fastparse.bytes` example: a UDP packet parser, which I made into the top-example for the`Byte Parsers` docs

- Moved `ScalaParse` doc under the `Example Parsers`, since top-level was getting a bit crowded

- Added a new documentation section under `Internals`, to mention how two-way round-trip parse -> re-serialization isn't supported by FastParse.

- Flipped the order of `Equivalent Byte Parsers` and `Byte-specific Parsers`, now that the primary example for byte parsers uses things like `BE.UInt16`

- Turn on Coursier as a SBT plugin to make ivy resolution faster

Review by @vovapolu

### DIFF
--- a/fastparse/shared/src/main/scala/fastparse/Api.scala
+++ b/fastparse/shared/src/main/scala/fastparse/Api.scala
@@ -92,7 +92,7 @@ class ByteApi() extends Api[Byte, Array[Byte]]() {
     if (seq.length == 1) parsers.Terminals.ElemLiteral(seq(0))
     else parsers.Terminals.Literal(seq)
 
-  object BytesParser {
+  object HexBytesParser {
     import all._
 
     val hexChars = HexUtils.hexChars
@@ -107,25 +107,32 @@ class ByteApi() extends Api[Byte, Array[Byte]]() {
     val bytes = all.P(byteSep ~ byte.rep(sep = byteSep)).map(_.toArray)
   }
 
-  def strToBytes(s: String): Array[Byte] = {
-    BytesParser.bytes.parse(s).get.value
+  /**
+    * Helper method to convert a space-separated string of hexidecimal bytes
+    * (e.g. "01 bc 21 04") into the corresponding Array[Byte]
+    */
+  def hexBytes(s: String): Array[Byte] = {
+    HexBytesParser.bytes.parse(s).get.value
   }
 }
 
-object byte extends ByteApi{
+object byte extends ByteApi {
   implicit def parserApi[T, V](p: T)
                               (implicit c: T => core.Parser[V, Byte, Array[Byte]]): ParserApi[V, Byte, Array[Byte]] =
     new ParserApiImpl[V, Byte, Array[Byte]](p)
 
+  /**
+    * Parses a two-byte word
+    */
   val Word16 = P( AnyByte.rep(exactly=2) )
+  /**
+    * Parses a four-byte word
+    */
   val Word32 = P( AnyByte.rep(exactly=4) )
+  /**
+    * Parses an eight-byte word
+    */
   val Word64 = P( AnyByte.rep(exactly=8) )
-}
-object byteNoApi extends ByteApi
-
-object ByteUtils {
-
-  import byte._
 
   trait ByteFormat {
     def wrapByteBuffer(byteSeq: ByteSeq): ByteBuffer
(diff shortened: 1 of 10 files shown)


fastparse/shared/src/test/scala/fastparse/BmpTests.scala
```diff
@@ -99,16 +98,14 @@ object BmpTests extends TestSuite {
...
-          .map(pixels => (header, pixels))
-    } ).map{
-      case (fileHeader, (bitmapHeader, pixels)) =>
-        Bmp(fileHeader, bitmapHeader, pixels.reverse)
-    }
+    val bmp = P(
+      for{
+        fileHeaderData <- fileHeader
+        headerData <- header
+        pixels <- bmpRow(headerData.infoPart.width, headerData.infoPart.bitsPerPixel)
```
vovapolu: Looks great! But it's a little big confusing that the ordering of for inner statements represents ordering of subparsers in parser.

fastparse/shared/src/main/scala/fastparse/Api.scala
```diff
@@ -107,8 +107,12 @@ class ByteApi() extends Api[Byte, Array[Byte]]() {
     val bytes = all.P(byteSep ~ byte.rep(sep = byteSep)).map(_.toArray)
   }
 
-  def strToBytes(s: String): Array[Byte] = {
-    BytesParser.bytes.parse(s).get.value
+  /**
+    * Helper method to convert a space-separated string of hexidecimal bytes
+    * (e.g. "01 bc 21 04") into the corresponding Array[Byte]
```
vovapolu: It can also handle notations like "0xAB" and isn't case sensitive (e.g "ab" == "AB" == "0xAb")

fastparse/shared/src/test/scala/fastparse/BmpTests.scala
lihaoyi: Sure, but I don't think it's any less confusing than the flatmap/nested-map + tuple-destructuring case though. Also, this matches roughly how some parser combinators work (e.g. all those in Haskell-land) so it's not entirely unprecedented

fastparse/shared/src/main/scala/fastparse/Api.scala
lihaoyi: do we want it to? I feel limiting it to one particular syntax may be better than leaving it "loose" like this.

readme/Readme.scalatex
```diff
@@ -1003,6 +1052,23 @@
...
+        @sect{Parsing Only}
+            @p
+                Fastparse, unlike libraries like @lnk("Scodec", "https://github.com/scodec/scodec")
+                or Python's @lnk("Construct", "https://pypi.python.org/pypi/construct"),
+                only allows you to define how to parse data from serialized
+                text/bytes into structured data, and not the other way around.
+
+            @p
+                It is probably possible to write a wrapper on top of FastParse
+                that will allow two-directional serialization, but such a thing
```
vovapolu: Btw, can you explain what "two-directional serialization" means? :) I'm just curious.

lihaoyi: Oh! I mean both `String => Foo` parsing, as well as `Foo => String` "unparsing". Many frameworks let you do that, but FastParse doesn't

fastparse/shared/src/main/scala/fastparse/Api.scala
vovapolu: Maybe we can make case sensitivity, but the "0x" prefix feature looks okay for me.

readme/Readme.scalatex
vovapolu: As I know, Scala Parser Combinators don't allow to do that too. And I always think Fastparse is some kind of replacement for slow but fp-style Scala library.

vovapolu:
The build is red, seems to be okay, but take a look at it.

lihaoyi:
Looks like flakes, but lemme restart those jobs see if that fixes it

lihaoyi:
I suspect (?) that there's a race condition in instantiating the `fastparse.byte` object, due to the new circular dependencies between `fastparse.byte` and `fastparse.ByteUtils`, that may be causing a deadlock on initialization. The tests are green, but lemme tweak it to try and avoid deadlock

vovapolu:
Why do you use force or squash commits? I can't see the progress of what you are changing.

lihaoyi:
Oh sorry, habit =P I guess I don't need to do that since github has a nice merge-squash button

The last changes are in https://github.com/lihaoyi/fastparse/pull/119/files#diff-72ca6d0349141f763e9e6a4350abb783R124, moving the contents of `ByteUtils` into the main `fastparse.bytes` object

lihaoyi:
Looks like we're solid green again https://travis-ci.org/lihaoyi/fastparse/builds/156476698

PR #122, opened 2016-09-02
Add a new example MidiParse

lihaoyi:
Also...

- Creates a simple `Midi` AST that it gets parsed into

- Creates a main method run via `sbt fastparseJVM/test:run`, that lets you load in a midi file, parse it into an AST, and interprets the AST to play music!

- Adds an description of MidiParse to the web-readme, including a live widget that lets users upload files, parses them, and uses MIDI.js/WebAudio to play them in the browser. Doesn't sound nearly as nice as the JVM-based version, but it'll do

- Made `.log()` less verbose: now if things fail and you call `.traced`, the second tracing parse won't go and duplicate all your logging output again

- Added all the various MIDI.js files to the repo. I couldn't find any convenient pre-packaged or CDNed version of it that worked (`build/MIDI.js` in their main repo doesn't seem to do what I want...) so I just added all the bits and pieces individually and linked them into the readme with script tags

- Added some simple unit tests to do basic sanity checking: that we can parse it, and some small sample of the midi-events are what we expect. Also added it to the perftest suite. Though honestly the files are large enough I'm mostly relying on "music sounds right" to give myself confidence in it's correctness 😛 

Review by @vovapolu. If the tests fail due to random stuff I'll fix 'em, but this is the first "large" parser I've written using `fastparse.byte` so I'd like you to give a pretty thorough review to make sure we're on the same page about how `fastparse.byte` is meant to be used


vovapolu:
Does that js midi player use .mp3 files to play midi notes? :D

lihaoyi:
@vovapolu yeah I think it does lol

lihaoyi:
It seems like there's a spec for it to be standard, but only chrome supports it lol so I guess MIDI.js polyfills it with mp3 files http://caniuse.com/#search=midi

lihaoyi:
Rebased on top of #124, let's see if tests pass

lihaoyi:
Probably worth trying to replace the heavy MIDI.js dependency with something lighter such as https://github.com/keithwhor/audiosynth before merging, to avoid bloating the size of the repo

lihaoyi:
@vovapolu Should be ready/easier to review now that #124 has merged. Also should be easier to review now that I've gotten rid of MIDI.js.

demo/src/main/scala/demo/DemoMain.scala
```diff
@@ -13,6 +14,158 @@ import scala.scalajs.js.typedarray.{ArrayBuffer, Uint8Array}
...
+          }
+          true
+        } else if (remaining.nonEmpty){
+
+          val milliSleep = midi.tickDiv match{
+            case fastparse.Midi.TickDiv.Metric(divisionsPerQuarterBeat) =>
+              val milliTempo = currentTempo / 1000
+              milliTempo * tillNext / divisionsPerQuarterBeat
+            case fastparse.Midi.TickDiv.TimeCode(fps, divs) =>
+              ???
```
vovapolu: is it going to be replaced?

demo/src/main/scala/demo/DemoMain.scala
```diff
@@ -13,6 +14,158 @@ import scala.scalajs.js.typedarray.{ArrayBuffer, Uint8Array}
...
+
+    val allTracks = midi.tracks.map(_.toList).map(TrackInfo(0, _))
+    // 120bpm, in microseconds-per-quarter-beat
+    var currentTempo = 1000000 * 120 / 60 / 4
+
+    def tick(): Unit = {
+
+      val logs = collection.mutable.Buffer.empty[String]
+      progressLogger(allTracks.map(_.track.length).mkString(", "))
+      while ({
```
vovapolu: This while true-false trick looks cool :) But is it really necessary? Maybe you could use some more scala-like technic?

fastparse/jvm/src/test/scala/fastparse/Main.scala
```diff
@@ -0,0 +1,105 @@
...
+    val synth = MidiSystem.getSynthesizer()
+    val channels = synth.getChannels()
+    synth.open()
+    case class TrackInfo(var savedTicks: Int, var track: List[(Int, Midi.TrackEvent)]){
+      def tillNext = track.head._1 - savedTicks
+    }
+    val allTracks = midi.tracks.map(_.toList).map(TrackInfo(0, _))
+
+    println("playing...")
+    while({
```
vovapolu: Seems like code-duplication, but I think it can cause more bulky code if you try to rewrite it.

fastparse/jvm/src/test/scala/fastparse/MidiTests.scala
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  def variousParses(bytes: Array[Byte]) = Seq(
+    MidiParse.midiParser.parse(bytes).get.value,
+    MidiParse.midiParser.parseIterator(bytes.grouped(1)).get.value,
+    MidiParse.midiParser.parseIterator(bytes.grouped(4)).get.value,
+    MidiParse.midiParser.parseIterator(bytes.grouped(16)).get.value,
+    MidiParse.midiParser.parseIterator(bytes.grouped(64)).get.value,
+    MidiParse.midiParser.parseIterator(bytes.grouped(256)).get.value,
+    MidiParse.midiParser.parseIterator(bytes.grouped(1024)).get.value
```
vovapolu: `for (size <- Seq(1, 4, 16, ...)) MidiParse.midiParser.parseIterator(bytes.grouped(size)).get.value`
Not like that? :)

readme/ByteParsers.scalatex
```diff
@@ -177,6 +178,81 @@
...
+                @li
+                    @code{midiEvent} and @code{metaEvent} show how to switch
+                    on a single byte upfront and use that to decide how to
+                    continue parsing a particular item, using
+                    @sect.ref{FlatMap}/@hl.scala{for}-comprehensions
+                @li
+                    How to use @sect.ref{Filter} to narrow down the sort of
+                    input a parser can parse, e.g. to narrow a
+                    @hl.scala{Int8: Parser[Byte]} into a parser that only
+                    accepts non-negative bytes
```
vovapolu: why doesn't midi format pack non-negative bytes into 7 bits considering all this mess with `varInt` and 24bits? :D

demo/src/main/scala/demo/DemoMain.scala
lihaoyi: Let's leave it `???` for now. I couldn't find any midi files on my machine that would trigger that code path, so I'm happy to leave it as "not implemented" for the time being. If someone complains about it not working, then I'll look into fixing it

demo/src/main/scala/demo/DemoMain.scala
lihaoyi: Well, it's not strictly necessary, but it's convenient. `do-while` doesn't work because the things you define inside the `do` block aren't available in the conditional when deciding whether or not to continue. In a more traditional language I would use `while(true)` and `break`, but `break` in Scala is funky (it throws exceptions!), and nobody really uses it, so I'd rather avoid it. I could re-write it as tail-recursive, but that makes the code a lot messier than a simple loop

fastparse/jvm/src/test/scala/fastparse/Main.scala
lihaoyi: Yeah it's code duplication, but the two implementations are sufficiently different to make re-use annoying

- One i running on JVM/MidiSystem, the other JS/AudioSynth
- One is a synchronous while-loop, the other is a semi-asynchronous while-loop-embedded-in-timer-callback
- The JVM side can handle a lot more of the midi file format than the JS side: different channels, pressure-thingies, etc.. JS just plays a single fixed "instrument"
- The JS side needs to do some initialization to pre-generate all the wave-forms before running, the JVM side doesn't

Overall, while I'm sure the bits in between could be shared, it would be pretty difficult to share the shareable bits while working around all the differences between the two implementations

fastparse/jvm/src/test/scala/fastparse/MidiTests.scala
lihaoyi: Ah yeah, I can do that =)

readme/ByteParsers.scalatex
lihaoyi: I don't understand your question? The `Int8.filter(_ >= 0)` thing is just the way many of the fields are specified, which separate from the `varInt` fields. Why some things are `varInt` and some things are `posInt8` and some things are 24-bit ints, I can only assume they had some reason (legacy? backwards-compatibility? performance? seek-ability?) for choosing so.

vovapolu: Never mind, midi-format just looks quite knotty for me

lihaoyi: Yeah, it is. I mostly used http://www.somascape.org/midi/tech/mfile.html#midi as a reference when writing the parser, so if you're wondering whether the parser is wrong or the format is just weird, it's possibly worth taking a slow read through that doc just to familiarize yourself with what the format looks like. It's not that long a format, really, just has some subtleties around the various number formats. It's only 12 pages so not *that* much "stuff" in it

lihaoyi:
@vovapolu I have removed the duplication in the midiparse iterator tests

PR #123, opened 2016-09-02
Added support for literal types.

milessabin:
Before I go any further with this, I'd like to know how you want to handle this feature being enabled/disabled ...


lihaoyi:
I'd be perfectly happy having be non-disable-able to be honest. ScalaParse is already a "sloppy" parser and accepts more than the Scalac parser does, and this shouldn't cause existing "good" code to parse in a different way AFAICT

milessabin:
Okey doke ... I've added a few more tests, and removed the failing tests which are now expected to pass.

lihaoyi:
Looks good to me, thanks @milessabin! This will go out in 0.4.0 after https://github.com/lihaoyi/fastparse/pull/122 lands and I'll publish a corresponding new version of Ammonite. Realistically probably in a week or so

milessabin:
Awesome! Thanks :-)

PR #124, opened 2016-09-03
Properly handle cuts in LHS of flatMap calls and Not/Lookahead/Filter

lihaoyi:
We were not properly propagating the `cut`s in the LHS of `flatMap`s; that meant that if there was a `cut` inside the left hand parser, it doesn't take effect if something fails in the RHS of the `flatMap` call, or if the `flatMap` succeeds but something fails later.

Also makes `!` and `&` negative/positive lookaheads behave as `NoCut`s: the whole point of those operators is to backtrack, so there isn't any conceivable use-case where someone would want to use them and have a cut disable-backtracking/drop-buffer when inside one of these operators.

Also properly propagates cuts in `foo.filter(pred)` when `foo` passes but `pred` fails.

Found these when investigating the build failure in https://github.com/lihaoyi/fastparse/pull/122. Possibly not the only issue, but it's something. This changes behavior, but if this makes any tests fail, I'll fix them.

Review by @vovapolu

### DIFF
--- a/fastparse/shared/src/main/scala/fastparse/core/Parsing.scala
+++ b/fastparse/shared/src/main/scala/fastparse/core/Parsing.scala
@@ -215,6 +215,7 @@ trait Mutable[+T, ElemType]{
    * this success, which prevents backtracking.
    */
   def cut: Boolean
+  def cut_=(c: Boolean): Unit
 }
 object Mutable{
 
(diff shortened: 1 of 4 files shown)


lihaoyi:
In theory, this should be all that is necessary to make https://github.com/lihaoyi/fastparse/pull/122 pass

PR #125, opened 2016-09-04
Polish and documentation work for fastparse.byte

lihaoyi:
This is a bit of a big/messy diff, that basically includes a lot of cleanup and tweaks and docs to help polish the API

- Optimize `[U]Int{8, 16, 32, 64}` to avoid allocating temporary arrays and byte buffers
- Add `Float32` and `Float64` parsers
- Swap the order of arguments in `repeatWithLength`, so it matches the order of parsing
- Make `hexBytes` more forgiving to whitespace, so it works with multi-line strings
- Remove `ByteSeq` alias since we already have `BS`
- Re-arrange the `ElemsX`/`ByteX`/`CharX` parsers; give the `ByteX`/`CharX` parsers toStrings
  that match their ByteX/CharX names
- Move `prettyBytes` into the main `fastparse.byte` API, document it
- Add `AnyElems`/`AnyBytes`/`AnyChars` parser, for quickly chomping arbitrary
  fixed-length input.
- Update existing parsers to make heavy use of `AnyBytes`, `WordN`, and `IntN` parsers
- Add a ton of minimal-example docs for all the byte specific parsers
- Add a new faux-struct parser to be the intro example for byte parsers, move UDP parser
  down into "example byte parsers" section


lihaoyi:
This is basically ready now the build is green; when #122 lands I'll squash everything and clean this up for review

vovapolu:
It's getting complicated :) There are 2 PRs and you pushed new commit here.

lihaoyi:
Oops, yeah lemme push to the other one

lihaoyi:
Ok, time to clean this up and now that #122 has landed

fastparse/shared/src/main/scala/fastparse/ByteUtils.scala
```diff
@@ -0,0 +1,199 @@
...
+                  contextRows: Int = 8) = {
+    val invalidIndices = markers.filter(
+      x => x < -1 /*-1 is special case*/ || x >= bytes.length
+    )
+    require(
+      invalidIndices.isEmpty,
+      s"Cannot show indices $markers outside input bounds 0 -> ${bytes.length}"
+    )
+
+    val maxIndexWidth = math.floor(math.max(0, math.log10(99))).toInt + 1
```
vovapolu: It equals to 2 :)

fastparse/shared/src/test/scala/fastparse/ByteTests.scala
```diff
@@ -266,37 +267,579 @@ object ByteTests extends TestSuite {
       }
     }
     'example{
+      'bs{
+        import fastparse.byte._
+
+        assert(BS(0x01, 0xff).toSeq == Array[Byte](0x01.toByte, 0xff.toByte).toSeq)
+        assert(BS("hello":_*).toSeq == Array[Byte](104, 101, 108, 108, 111).toSeq)
```
vovapolu: Hmm, I missed this conversion from char* to bytes*. Can you explain it more detailed?

fastparse/shared/src/test/scala/fastparse/ByteTests.scala
```diff
@@ -266,37 +267,579 @@ object ByteTests extends TestSuite {
...
+      }
+      'hexBytes{
+        import fastparse.byte._
+
+        assert(hexBytes("01 ff").toSeq == Array[Byte](0x01.toByte, 0xff.toByte).toSeq)
```
vovapolu: Looks like very poor test-set ;)

vovapolu: You can take good example from classparse test.

fastparse/shared/src/test/scala/fastparse/ByteTests.scala
```diff
@@ -266,37 +267,579 @@ object ByteTests extends TestSuite {
...
+          """      0  1  2  3  4  5  6  7  8  9  10 11 12 13 14 15
+            |
+            |0     54 68 65 20 71 75 69 63 6b 20 62 72 6f 77 6e 20
+            |16    66 6f 78 20 6a 75 6d 70 73 20 6f 76 65 72 20 74
+            |32    68 65 20 6c 61 7a 79 20 64 6f 67 2e 20 53 68 65
+            |48    20 73 65 6c 6c 73 20 73 65 61 20 73 68 65 6c 6c
+            |64    73 20 6f 6e 20 74 68 65 20 73 65 61 20 73 68 6f
+            |80    72 65 20 62 75 74 20 74 68 65 20 73 68 65 6c 6c
+            |96    73 20 73 68 65 20 73 65 6c 6c 73 20 73 68 65 20
+            |112   73 65 6c 6c 73 20 6e 6f 20 6d 6f 72 65 2e 20 50
```
vovapolu: Maybe it's worth to add option of 0x-notation indices, cause it's used in other hex editors, as I remember.

fastparse/shared/src/main/scala/fastparse/ByteUtils.scala
lihaoyi: lol nice catch, yeah that should be `bytes.length`

fastparse/shared/src/test/scala/fastparse/ByteTests.scala
lihaoyi: I made `BS` take anything of type `T: Numeric`; that includes `Byte`, `Short`, `Int` (that's why the `0xff` notation works; it didn't work earlier when it only took `Byte`) and `Char`, which is also defined as `: Numeric`. You can after all add and subtract `Char`s from each other, so in Scala there's an implicit `Numeric[Char]` available. 

Thus, we can take a `String`, splat it into the `BS` function into it's individual chars, and then (at least for ASCII chars) they'll get converted to the correct bytes.

Chars *outside* the byte-range of 0-255 will get truncated to be within 0-255, so all the unicode characters won't work, but ASCII works fine

fastparse/shared/src/test/scala/fastparse/ByteTests.scala
lihaoyi: lol good point.

fastparse/shared/src/test/scala/fastparse/ByteTests.scala
lihaoyi: Hmm we could, but I'd push back a little and say let's punt on that for now, maybe open an issue, and consider implementing that after 0.4.0 is out.

vovapolu: Yeah, okay. I don't think it's very important right now, but as you start this thing with byte pretty formatting, you can make it more rich and with various options.

vovapolu:
@lihaoyi New tests are wrong, and there is also some big debug output for new midi parser that distracts a little.

lihaoyi:
Oh wat, I thought I had things running green before pushing. Lemme take a look

lihaoyi:
Looks like it was a JS/JVM incompatibility; I only tested it on JVM locally. Will fix

lihaoyi:
Turns out `math.log10(1000)` returns `3.0` in Scala-JVM and `2.9999999999999996` in Scala.js, and this was causing the calculation of number-width to be wrong. I'll just replace it with a `bytes.length.toString.length`

fastparse/shared/src/main/scala/fastparse/ByteUtils.scala
```diff
@@ -21,7 +21,7 @@ object ByteUtils{
       s"Cannot show indices $markers outside input bounds 0 -> ${bytes.length}"
     )
 
-    val maxIndexWidth = math.floor(math.max(0, math.log10(bytes.length))).toInt + 1
+    val maxIndexWidth = bytes.length.toString.length
```
vovapolu: lol, nice solution :D

PR #127, opened 2016-09-09
Convert fastparse.byte to use scodec.bits.ByteVector instead of Array[Byte]

lihaoyi:
This is an experiment to see what it will look like if we use `ByteVector` instead of `Array[Byte]`. Major changes:

- `hexBytes` is gone, replaced by Scodec's `hex""` interpolator

- FastParse now depends on `scodec-bits`.

- `scodec.bits.ByteVector` is now aliased to `fastparse.byte.Bytes`, usually used as just `Bytes`, for conciseness

- Implicit conversion between `Array[Byte] => Parser[Unit, Byte, Array[Byte]]` is now gone, as if we make it `Bytes => Parser[Unit, Byte, Bytes]`, then using the `|` operator on two `Bytes` values within a parser behaves surprisingly: it bit-wise `|`s the two `Bytes` values together rather than constructing a parser that can parse either.

- `BS` now constructs `Parser[Unit]` instead of constructing `Array[Byte]` and relying on the implicit to convert it to a parser, since the implicit is gone. It is also overloaded to take `Bytes` instead of the literal byte-values, so if you want to convert `Bytes` to a `Parser[Unit]` you just wrap it in `BS`

- Add example unit-tests for `Bytes`/`ByteVector`, and use as documentation on how to use it

- Update all existing parsers in the repo to make use of `Bytes`, `hex""` and the rest of the new API. All the old `.toSeq` calls before comparing arrays are now gone, since `==` works.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -63,7 +63,8 @@ lazy val utils = crossProject
     unmanagedSourceDirectories in Compile ++= {
       if (scalaVersion.value startsWith "2.10.") Seq(baseDirectory.value / ".."/"shared"/"src"/ "main" / "scala-2.10")
       else Seq(baseDirectory.value / ".."/"shared" / "src"/"main" / "scala-2.11")
-    }
+    },
+    libraryDependencies += "org.scodec" %%% "scodec-bits" % "1.1.0"
   )
 lazy val utilsJS = utils.js
 lazy val utilsJVM= utils.jvm
@@ -101,7 +102,8 @@ lazy val fastparse = crossProject
         """.stripMargin
       IO.write(file, output)
       Seq(file)
-    }
+    },
+    libraryDependencies += "org.scodec" %%% "scodec-bits" % "1.1.0"
   )
   // In order to make the midi-parser-test in fastparseJVM/test:run work
   .jvmSettings(fork in (Test, run) := true)
--- a/fastparse/shared/src/test/scala/fastparse/BmpTests.scala
+++ b/fastparse/shared/src/test/scala/fastparse/BmpTests.scala
@@ -2,7 +2,6 @@ package fastparse
 
 import scala.collection.mutable.ArrayBuffer
 import fastparse.byte._
-import scodec.bits.ByteVector
 import utest._
 
 /*
@@ -30,7 +29,7 @@ object BmpTests extends TestSuite {
 
       case class BitmapInfoHeader(infoPart: BitmapInfoHeaderPart) extends BitmapHeader(infoPart)
 
-      case class Pixel(colors: ByteVector)
+      case class Pixel(colors: Bytes)
 
       case class Bmp(fileHeader: FileHeader, bitmapHeader: BitmapHeader, pixels: Seq[Seq[Pixel]])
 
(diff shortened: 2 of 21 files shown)


build.sbt
```diff
@@ -101,7 +102,8 @@ lazy val fastparse = crossProject
         """.stripMargin
       IO.write(file, output)
       Seq(file)
-    }
+    },
+    libraryDependencies += "org.scodec" %%% "scodec-bits" % "1.1.0"
```
lihaoyi: Before publishing 0.4.0 I'd probably want to break out the `scodec-bits` dependency into a separate `fastparse-byte` subproject, so the core `fastparse` project doesn't pick up unnecessary dependencies

lihaoyi:
Woo build is green! @vovapolu this is ready for you to review. What do you think of this change in general? Good/Bad?

CC @mpilquist in case you want to poke around the diff to see how I'm using `scodec.bits.ByteVector`

lihaoyi:
FWIW I ran `perftests.binary.MidiParse` locally and as far as I can tell there's no big difference in the runtimes

vovapolu:
Well, if there's no big difference, so why we need to change anything?

lihaoyi:
A few things

- Having the underlying data-model be immutable is nice, v.s. "here's a mutable Array, but please please please don't write to it". When you look at our original code, we had nice immutable case classes with these mutable `Array[Byte]`s as fields everywhere, which kind of defeats the point of immutability.

- Having a good `.toString`, `==` makes a lot of things easier: e.g. testing and debugging. Having all our case classes have non-working `==` due to some `Array[Byte]` nested deep inside was annoying.

- Get a bunch of nice things for free, e.g. the `hex""` interpolator, that we no longer need to implement ourselves

- Convenient constructors from hex, base64, `ByteBuffer`, fast concats and appends, and many other useful things. Many more than if we tried to implement our own immutable byte array

Overall, adding a dependency on a third-party library is a negative, but I think this one is lightweight enough and non-changing enough that the benefits are worth it.

fastparse/shared/src/test/scala/fastparse/BmpTests.scala
```diff
@@ -114,31 +114,31 @@ object BmpTests extends TestSuite {
   val tests = TestSuite {
     def compareBmps(bmp1: Bmp, bmp2: Bmp): Boolean ={
```
vovapolu: It's not needed now, cause case class has the same `equals` method.

fastparse/shared/src/test/scala/fastparse/ByteTests.scala
```diff
@@ -267,29 +277,115 @@ object ByteTests extends TestSuite {
...
-        ).toSeq)
+        assert(
+          hex"01 ff" == Bytes.fromHex("01 ff").get,
+          hex"""
+            de ad be ef
+            ca fe ba be
+          """ == Bytes.fromHex("""
+            de ad be ef
+            ca fe ba be
+          """).get
```
vovapolu: Maybe it's worth to test `hex""` and `fromHex` independently? Are we sure that they both work correctly?

fastparse/shared/src/test/scala/fastparse/BmpTests.scala
lihaoyi: Woohoo

fastparse/shared/src/test/scala/fastparse/ByteTests.scala
lihaoyi: They're lightly tested in scodec-bits, though their test suite isn't great. I'll add a few cases

mpilquist: @lihaoyi PRs happily accepted for improving the test suite in scodec-bits. :)

lihaoyi: @mpilquist understood 😛

lihaoyi:
bump @vovapolu

lihaoyi:
If there's no other feedback I'm moving forward with this

vovapolu:
@lihaoyi Sorry for such late, it was a busy week.

lihaoyi:
@vovapolu sure no problem =) I'm working on polishing the API and cleaning up parts of the codebase so we can release 0.4.0. Hoping to get it out this weekend, so people can start using your summer's worth of work

PR #134, opened 2016-11-09
Add ElemsIn intrinsic

edmundnoble:
`ElemWhile` requires enumerating every character there is on initialization, which is slow (at least in JS) but `ElemIn` and `.rep` are too slow at parse-time for some applications, for the same reason there's an `ElemWhile` in the first place. Thus, there is a need for a primitive combinator combining `ElemWhile` and `ElemIn`.


lihaoyi:
I think the correct thing to do here is to let people pass in a flag to `ElemIn` `ElemWhile` or `ElemPred` to control or not we pre-compute things, rather than a whole new element. Fits a bit better, since precompute-or-not applies to all those existing intrinsics equally rather than being it's own special-thing

lihaoyi:
Use case should be satisfied by 04b5ac2f688c1f937110158c7dc5b7870cdb02bb

PR #136, opened 2016-12-17
Parsing support of Vectors[_] or others sequences converted to Vector

math85360:
Hi,

For my own purpose, I've tried to implement Vector input parsing support.

If anybody needs it, you can use any Sequence and convert it to Vector to use as input for this parsing support.

See fastparse/shared/src/main/scala/fastparse/sequence/VectorApi.scala for more details to use it.

Thanks ! 😃


vovapolu:
@math85360 Looks very cool :) I'll take a look when I have a free time.

vovapolu:
@math85360 Can you add several unit tests for this? It will help to understand the new features and  it's just good to have unit tests :)

vovapolu:
Yep, looks good! After adding unit-tests, I will read it again (maybe add one more example of usage :) ) and merge it.

lihaoyi:
Taking a look at this...

fastparse/shared/src/main/scala/fastparse/sequence/VectorApi.scala
```diff
@@ -0,0 +1,78 @@
...
+class VectorApi[A]()(implicit evidence: ClassTag[A]) extends SeqApi[A, Vector[A]](implicitly, VectorReprOps[A]) {
+  val AnyElem = parsers.Terminals.AnyElem[A, Vector[A]]("AnyElem")
+  def AnyElem(count: Int) = AnyElems[A, Vector[A]]("AnyElems", count)
+  def ElemPred(pred: A => Boolean): P0 = SeqItemParsers.Intrinsics.ElemPred[A, Vector[A]]("ElemPred", pred)
+  def ElemIn(items: Seq[A]*): P0 = SeqItemParsers.Intrinsics.ElemIn[A, Vector[A]]("ElemIn", items: _*)
+  def ElemsWhile(pred: A => Boolean, min: Int = 1) = SeqItemParsers.Intrinsics.ElemsWhile[A, Vector[A]]("ElemsWhile", pred, min)
+}
+
+/**
+ * This class is used to parse a Vector of items
```
lihaoyi: Rather than including this as a doc-comment, let's include it in the generated doc-site as a separate section, below `Byte Parsers`: https://github.com/lihaoyi/fastparse/blob/master/readme/Readme.scalatex#L24

Perhaps we could call it `Token Parsers` or similar. That should then include code examples extracted from your unit tests, so e.g. instead of copy-pasting the example into the docs we'd tidy up some of the examples you already have in `SequenceTests.scala` and embed them using `@hl.ref`.

We should follow the rough layout of the existing `Byte Parsers` section (http://www.lihaoyi.com/fastparse/#ByteParsers):

- Simple, motivating example

- Any operators that are specific to token parsing (e.g. `F{}`)

- A few of the token parsers that are identical to the string parsers. No need to have as exhaustive a list as http://www.lihaoyi.com/fastparse/#EquivalentByteParsers, but we should have at least some comparisons

- At least one slightly-meatier, more-realistic token parser to better motivate what these token parsers are used for, and provide an example for how the parsers look like when they grow beyond trivial cases.

fastparse/shared/src/main/scala/fastparse/sequence/VectorReprOps.scala
```diff
@@ -0,0 +1,34 @@
+package fastparse.sequence
+import fastparse.utils._
+import scala.reflect.ClassTag
+
+case class VectorReprOps[T]()(implicit evidence: ClassTag[T]) extends fastparse.utils.ReprOps[T, Vector[T]] {
+  type U = Vector[T]
+  val U = Vector
```
lihaoyi: This `val` does not appear to be used much; we should probably remove it and directly use `Vector` unless it's common enough to justify it's existence

fastparse/shared/src/main/scala/fastparse/sequence/VectorReprOps.scala
```diff
@@ -0,0 +1,34 @@
+package fastparse.sequence
+import fastparse.utils._
+import scala.reflect.ClassTag
+
+case class VectorReprOps[T]()(implicit evidence: ClassTag[T]) extends fastparse.utils.ReprOps[T, Vector[T]] {
+  type U = Vector[T]
```
lihaoyi: `U` is a bit of an odd name here. Could we call it `type VecT`? Slightly longer but much more clear

fastparse/shared/src/main/scala/fastparse/sequence/VectorReprOps.scala
```diff
@@ -0,0 +1,34 @@
...
+  override def length(input: U) = input.length
+
+  override def fromArray(input: Array[T]) = input.toVector
+  override def fromSeq(input: Seq[T]) = input.toVector
+  override def fromSingle(input: T) = U(input)
+  override def toArray(input: U) = input.toArray
+  override def flatten(input: Seq[U]) = input.flatten.toVector
+  override def prettyPrint(input: U) = input.mkString("[", ", ", "]")
+  override def literalize(input: U): String = prettyPrint(input)
+  override def errorMessage(input: fastparse.utils.ParserInput[T, U], expected: String, idx: Int): String = {
```
lihaoyi: It's not clear to me what this `errorMessage` implementation is doing; let's have one unit test that checks the error message to make sure it's what you expect, and that readers of the code could use to see what the final output of this code is meant to be

fastparse/shared/src/main/scala/fastparse/sequence/SeqApi.scala
```diff
@@ -0,0 +1,62 @@
...
+
+import scala.reflect.ClassTag
+import fastparse.core
+import fastparse.parsers
+
+/**
+ * This is a trait applicable on potentially any sequences but Vector or Array
+ * are typically usable in real world.
+ */
+abstract class SeqApi[Elem, Repr <: Seq[Elem]](ct: ClassTag[Elem], reprOps: fastparse.utils.ReprOps[Elem, Repr]) {
```
math85360: Yes for IndexedSeq[Elem].
On scala.List, it needs to create tests to see if it's really slow.
I think the usage case is when we've a large flat sequence of "primitives" to reorder and reduce, so in my head, access to List is slower than Vector (but maybe I'm wrong) ?

fastparse/shared/src/main/scala/fastparse/sequence/SeqApi.scala
```diff
@@ -0,0 +1,62 @@
...
+  implicit val implicitReprOps = reprOps
+
+  protected[this] implicit val implicitClassTag = ct
+  type ParserInput = fastparse.utils.ParserInput[Elem, Repr]
+  type IndexParserInput = fastparse.utils.IndexedParserInput[Elem, Repr]
+  type IteratorParserInput = fastparse.utils.IteratorParserInput[Elem, Repr]
```
math85360: Because there is `elemSetHelper: ElemSetHelper[Elem]` in Api.
So I can't convert any object into an Int or toLowerCase.

math85360:
Thank you for this complete review and all your comments !
Sorry for the formatting, have you a scalariform-settings file or equivalent ?

I don't know if I'm the only one with this usage case but maybe other people can have similar needs.
And I love fastparse library, so when I think I can help a little, I do it.

In my usage, I parse a big string into some primitives and next I need to reorder them and reduce into smarter objects.
At the beginning, I try with Pattern Matching and a lot of recursive extractors.
But it was a nightmare, and when I see it could be possible with some new classes for fastparse, I do it.

lihaoyi:
I don't have an automatic formatting config right now, so we'll just have to hash out the formatting back and forth manually for now. 

For such a large contribution, we need to make sure we maintain the level of quality and usability that are probably why you're using FastParse in the first place. That means a good set of motivating examples, verbose docs, and heavy testing (all three are related). 

That takes a good amount of work to make happen; if you're willing to do it, I'll help walk you through it. Most of the necessary steps have already been given in my earlier review, with regard to what exactly needs to happen to bring this up to the same standard as the rest of the codebase.

If you're not free to push it through that's fine too: we'll leave this here for future reference. In future, if someone is interested in it they can make use of your code directly, or they may themselves decide to put in the work to push it through.

tannerezell:
Hoping to see this merged, looks very useful!

lihaoyi:
Yeah I hope to see this merged too; someone must put in the work to make it mergeable tho 😛

math85360:
I'm fixing many thing (taking account your comments).
And today, I'm working on an example : an arithmetic simplifier.
I think something will be commited in few hours !

lihaoyi:
Out of date due to 2.0.0

PR #138, opened 2017-01-13
pythonparse: Parse unary operations and negative numbers

jeremydhoon:
### Summary

This PR allows `pythonparse` to recognize two previously-ignored constructs:

- Negative number literals.
- Unary operations `UAdd`, `USub`, and `Invert`.

### Details

Previously, `pythonparse` discarded these operators, rather than surfacing them as part of the constructed `Ast`. The logic linked below was responsible for discarding unary operators:

https://github.com/lihaoyi/fastparse/compare/master...jeremydhoon:jhoon-unaryop-invert?expand=1#diff-7fe7f240cf772c974bc24a8ecc6bd120L91

To capture these operators properly, the PR makes three changes:
- When parsing numbers, we capture a preceding sign operator and use it in computing the number's Scala value.
- When parsing `factor`s, we capture preceding unary operators, wrapping the subsequent expression with a `UnaryOp` if found.
- Before attempting to parse a `factor`, we first attempt to parse a `NUMBER`. This ensures numbers will parse as `Num(-1)` rather than `UnaryOp(USub, Num(1))`.

### Testing

I added a handful of regression tests initially, and then added a number unit tests as well. I'd be more than happy shift tests around to their proper files if this was not the correct arrangement.

When writing tests, I tried to ensure that the expected parse tree matched that returned by Python's `ast` module with `ast.dump(ast.parse("..."))`.

### DIFF
--- a/pythonparse/shared/src/main/scala/pythonparse/Ast.scala
+++ b/pythonparse/shared/src/main/scala/pythonparse/Ast.scala
@@ -146,7 +146,7 @@ object Ast{
     case object Invert extends unaryop
     case object Not extends unaryop
     case object UAdd extends unaryop
-    case object USubextends extends unaryop
+    case object USub extends unaryop
   }
   sealed trait cmpop
   object cmpop{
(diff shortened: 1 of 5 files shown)


lihaoyi:
Fix the one typeo I pointed out and I'll merge this in

jeremydhoon:
Thanks! For some reason I'm not seeing your comment pointing out the typo. I'm embarrassed to say I re-read the PR and didn't spot a typo myself, either.

Mind dropping another comment to point me to the right spot?

lihaoyi:
Subextends should just be Sub

lihaoyi:
Yep, compatibility with these example parsers is pretty YOLO right now. 

Thanks, this looks good!

PR #139, opened 2017-01-13
be Scala 2.13 friendly

SethTisue:
for the Scala 2.13 community build (yes, there is one already)


lihaoyi:
Huh seems to have broken the build for 2.12; seems we're including multiple `scala-2.1X/` folders and it's resulting in duplicate `Compat` objects

SethTisue:
fixed version pushed, let's see if Travis agrees

SethTisue:
Travis agrees.

lihaoyi:
Looks good to me thanks!

PR #140, opened 2017-01-13
pythonparse: Do not parse parenthetical expressions as single-item tuples

jeremydhoon:
### Summary

When parsing an expression such as `(a + b) * (c - d)`, the parenthesized terms should be treated as scalar expressions. The current behavior in `pythonparse` is to treat these terms as single-item tuples.

### Details

The PR adds a different parser for the contents of tuples, which ensure the presence of at least one `test` followed by a comma in order to consider the expression a tuple. To capture the scalar expression case, the PR adds a case for `test` to the parser for parenthesized terms under `atom`.

### Testing

The PR adds two brief unit test cases. I verified that the expected parses match the results produced by `ast.dump(ast.parse('...'))` in Python.

### DIFF
--- a/pythonparse/shared/src/main/scala/pythonparse/Expressions.scala
+++ b/pythonparse/shared/src/main/scala/pythonparse/Expressions.scala
@@ -105,7 +105,7 @@ object Expressions {
       empty_tuple  |
       empty_list |
       empty_dict |
-      "(" ~ (yield_expr | generator | tuple) ~ ")" |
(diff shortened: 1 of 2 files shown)


lihaoyi:
This looks good. The parsing of things like `(((a))` is now exponential, but we have enough random backtracking in the parser that there's probably plenty of other exponential places in there, so we can deal with those later if/when they cause problems.

Thanks again! Keep these fixes coming 👍

PR #145, opened 2017-03-18
Add support for trailing commas

dwijnand:
Ref scala/scala-dev#299.

@lihaoyi I need a hand failing the neg tests, could you have a look, please?

### DIFF
--- a/scalaparse/shared/src/main/scala/scalaparse/syntax/Literals.scala
+++ b/scalaparse/shared/src/main/scala/scalaparse/syntax/Literals.scala
@@ -32,6 +32,7 @@ trait Literals { l =>
     val ConsumeComments = P( (Basic.WSChars.? ~ Literals.Comment ~ Basic.WSChars.? ~ Basic.Newline).rep )
     P( NoCut( WS ~ Basic.Newline.? ~ ConsumeComments ~ NotNewline) )
   }
+  val TrailingComma: P0 = ("," ~ Basic.Newline).?
   def Pattern: P0
   object Literals{
     import Basic._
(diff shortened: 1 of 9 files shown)


dwijnand:
@lihaoyi This should go green. Could you have a look and let me know what you think, please?

And/or any feedback from @fizzy33 :)

scalaparse/shared/src/main/scala/scalaparse/syntax/Literals.scala
```diff
@@ -32,6 +32,7 @@ trait Literals { l =>
     val ConsumeComments = P( (Basic.WSChars.? ~ Literals.Comment ~ Basic.WSChars.? ~ Basic.Newline).rep )
     P( NoCut( WS ~ Basic.Newline.? ~ ConsumeComments ~ NotNewline) )
   }
+  val TrailingComma: P0 = ("," ~ WS.rep ~ Basic.Newline).?
```
dwijnand: @lihaoyi I'm not sure if I should use WS vs WL0 vs WL, cut vs NoCut, etc here. I'm very confused :)

lihaoyi:
Seems to hang on one of the failure tests, looking into it

scalaparse/shared/src/main/scala/scalaparse/syntax/Literals.scala
lihaoyi: This `.rep` is unnecessary, since the `WS` internally already loops 0-or-more times, and the fact that `WS` can succeed when consuming 0 characters means the surrounding `.rep` can succeed infinitely without making forward progress. I have removed it, and all your new tests cases still seem to pass, so that's probably fine

dwijnand:
Thanks for taking care of this @lihaoyi.

PR #152, opened 2017-03-26
Rewrite the XML parser based on the scalac one

allanrenucci:
The actual xml parser does not match the one from scalac. This parser is based on the one from [scalac](https://github.com/scala/scala/blob/2.12.x/src/compiler/scala/tools/nsc/ast/parser/MarkupParsers.scala).

### DIFF
--- a/scalaparse/shared/src/main/scala/scalaparse/Xml.scala
+++ b/scalaparse/shared/src/main/scala/scalaparse/Xml.scala
@@ -8,123 +8,110 @@ import scala.language.implicitConversions
 
 trait Xml extends Core {
   def Patterns: P0
-  val XmlExpr = P( WL ~ Xml.XmlContent ~ (WL ~ Xml.Element).rep )
+  val XmlExpr = P( WL ~ Xml.XmlContent.rep(min = 1, sep = WL.?) )
   val XmlPattern = P( WL ~ Xml.ElemPattern )
 
-  private[this] object Xml{
-    val BaseChar = P(CharIn(
-      '\u0041' to '\u005A', '\u0061' to '\u007A', '\u00C0' to '\u00D6', '\u00D8' to '\u00F6',
-      '\u00F8' to '\u00FF', '\u0100' to '\u0131', '\u0134' to '\u013E', '\u0141' to '\u0148', 
-      '\u014A' to '\u017E', '\u0180' to '\u01C3', '\u01CD' to '\u01F0', '\u01F4' to '\u01F5', 
-      '\u01FA' to '\u0217', '\u0250' to '\u02A8', '\u02BB' to '\u02C1',  "\u0386", 
-      '\u0388' to '\u038A', "\u038C", '\u038E' to '\u03A1', '\u03A3' to '\u03CE', 
-      '\u03D0' to '\u03D6', "\u03DA", "\u03DC",  "\u03DE",  "\u03E0", '\u03E2' to '\u03F3', 
-      '\u0401' to '\u040C', '\u040E' to '\u044F', '\u0451' to '\u045C', '\u045E' to '\u0481', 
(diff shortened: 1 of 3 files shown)


scalaparse/shared/src/main/scala/scalaparse/Xml.scala
```diff
@@ -11,120 +11,108 @@ trait Xml extends Core {
...
+    val Name = P( NameStart ~ NameChar.rep )
+    val NameStart  = P( CharPred(isNameStart) )
+    val NameChar = P( CharPred(isNameChar) )
+
+    val ElemPattern: P0 = P( TagPHeader ~/ ("/>" | ">" ~/ ContentP ~/ ETag ) )
+    val TagPHeader = P( "<" ~ TagName ~ WL.?  )
+
+    val ContentP: P0 = P( (ElemPattern | CharDataP | ScalaPatterns).rep )
     val ScalaPatterns = P( "{" ~ Patterns ~ WL ~ "}" )
+    val CharDataP = P( "&" ~ CharData.? | CharData ) // matches weirdness of scalac parser on xml reference.
```
allanrenucci: I think this is a bug in scalac. A better solution might be to fix the scalac parser instead

scalaparse/shared/src/main/scala/scalaparse/Xml.scala
```diff
@@ -11,120 +11,108 @@ trait Xml extends Core {
...
+
+    //================================================================================
+    // From `scala.xml.parsing.TokenTests`
+    //================================================================================
```
allanrenucci: Should I add a dependency on `scala-xml` instead?

scalaparse/shared/src/test/scala/scalaparse/unit/FailureTests.scala
```diff
@@ -835,6 +835,16 @@ object FailureTests extends TestSuite{
         expected = """ "\"\"\"" | StringChars | Interp | NonTripleQuoteChar """,
         found = ""
       )
+      * - checkNeg(
+        """
+          |object System {
+          |  e match { case <xml:unparsed><</xml:unparsed> => }
```
allanrenucci: This test fails on 2.12 but not on 2.11 and 2.10.
```scala
expected  = CharsWhile(<function1>) | XmlPattern | Thingy |...
```
and
```scala
parsedExpected = CharsWhile(scalaparse.syntax.Identifiers$$$Lambda$5210/206499243@72995185) | XmlPattern | Thingy | ...
```
`toString` on lambda is different between 2.11 and 2.12

scalaparse/shared/src/test/scala/scalaparse/unit/SuccessTests.scala
```diff
@@ -1169,6 +1169,45 @@ object SuccessTests extends TestSuite{
       """.stripMargin
     )
     * - check(
+      """object Unparsed {
+        |    <xml:unparsed></xml:unparsed>
+        |    <xml:unparsed foo=""></xml:unparsed>
+        |    <xml:unparsed><</xml:unparsed>
```
lihaoyi: What on earth...

scalaparse/shared/src/main/scala/scalaparse/Xml.scala
lihaoyi: No, this is fine. Copy-pasting 10 lines of code beats a huge binary dependency

lihaoyi:
@olafurph @xeno-by: you guys are using a forked old-version of this XML parser in scala-meta, so if you want to bring it up to date you could fork it again

allanrenucci:
> @olafurph @xeno-by: you guys are using a forked old-version of this XML parser in scala-meta, so if you want to bring it up to date you could fork it again

@lihaoyi The latest version is in scalameta. If you want, I can open a new PR to bring your version up to date.

PR #153, opened 2017-04-24
Python 3.x parser

MartinWeindel:
This pull request provides a complete Python 3 parser (up to 3.5) as a new subproject.
The structure has been forked from the existing Python 2 parser subproject. Unit tests and project tests have been extended or rewritten accordingly.

The ProjectTests (i.e. python3parse/jvm/src/test/scala/python3parse/ProjectTests.scala) needs a python3 installation. Following projects from github have been successfully included for these tests:
- https://github.com/dropbox/changes
- https://github.com/python/typeshed
- https://github.com/pallets/flask
- https://github.com/zulip/zulip
- https://github.com/ansible/ansible
- https://github.com/kennethreitz/requests


lihaoyi:
I backed this out, due to problems in the PR

Using a global `mutable.Stack` in a parser isn't acceptable. That causes parallel parses to fail nondeterministically. You have to rewrite it using the same ExprCtx/FreeCtx/StmtCtx pattern we're using in Scalaparse

PR #154, opened 2017-05-13
More closely match scalac XML parser behavior

allanrenucci:


scalaparse/shared/src/main/scala/scalaparse/Xml.scala
```diff
@@ -59,9 +58,9 @@ trait Xml extends Core {
     val CharQ  = P( !"\"" ~ Char1 )
     val CharA  = P( !"'" ~ Char1 )
 
-    val Name      = P( NameStart ~ NameChar.rep )
-    val NameStart = P( CharPred(isNameStart) ).opaque("NameStart")
-    val NameChar  = P( CharPred(isNameChar) ).opaque("NameChar")
+    val Name: P0  = P( NameStart ~ NameChar.rep ).!.filter(_.last != ':').opaque("Name").map(_ => Unit) // discard result
```
allanrenucci: Is there a better way to do this?

PR #163, opened 2017-08-13
Reorganize build for easier cross-publish

olafurpg:
`+modules/publishSigned` will no longer work since native projects only cross-build against 2.11. sbt-doge solves this by making `very publishSigned` only run publishSigned for the defined crossScalaVersions in each project. However, the `modules` aggregate was messing up something so that the native modules were still being built against 2.10 with `very compile`. Removing `modules` fixed the problem.

While trying to debug the problems, I some changes to clean up build.sbt to better understand what was happening.


olafurpg:
I just released 0.4.4 with native support. With the latest commit in this PR, it's possible to publish/test/compile all modules for all scala version with `very <command>`.

I haven't pushed the tag yet or updated the docs. I propose we do that once this is merged.

olafurpg:
btw, sbt-doge is merged into sbt 1.0 so that `+` will work the same way as `very` does.

olafurpg:
Will go ahead and merge this to tag the release and update the docs. Please let me know if I messed up something 😅

PR #166, opened 2017-09-13
upgrade to latest utest (0.5.3)

SethTisue:
this is needed for the Scala community build; fansi requires the
latest utest now, so anything else using utest must be
source-compatible with the latest utest too


SethTisue:
the Travis failures look like they might be unrelated... the failing tests are ones that depend on external repositories, I think?

lihaoyi:
Yeah looks fine. Those are probably failing because some external codebase's code changed and some bits of it no longer parse. We really should pin the git versions of the code we pull in for those test suites, but i have not had time to do so. Feel free to send a PR to do that if you want to make this green

PR #173, opened 2017-10-26
Use a shallow clone in scalaparse ProjectTests

retronym:
I noticed that fastparse/target/repos is a disk hog in the community build.

This commit changes the checkout of third party repos by test cases to use a shallow clone,
as is already done in classparse/jvm/src/test/scala/classparse/ProjectTests.scala

References https://github.com/scala/community-builds/issues/617


lihaoyi:
Looks good to me thanks!

PR #178, opened 2018-01-30
add collect method

mberndt123:
Hey,

It seems to me that if parsers support `filter` and `map`, they should also support `collect`, which is what this PR adds. Please consider merging it.


lihaoyi:
Out of date due to 2.0.0

PR #179, opened 2018-01-31
change the order of the type parameters of core.Parser…

### LINKED ISSUE #176: Change order of core.Parser's type parameters to work with -Ypartial-unification
Hey there,

One of the more recent developments in the Scala language is the `-Ypartial-unification` flag, which essentially allows you to treat types with multiple type parameters as if they were curried. 
You could for instance declare a method like this:`def f[F[_], A](fa: F[A])`. If you pass an `Either[Int, String]` to `f`, it won't compile, because `Either` is a binary type constructor and `F` is unary. `-Ypartial-unification` fixes that. With this flag, scalac will infer that `F[X] = Either[Int, X]` and `A = String`. But if you pass a `core.Parser[Foo, String, Char]` type, it will infer `F[X] = Parser[Foo, String, X]` and `A = Char`. That's bad, because usually you'd want that to be `F[X] = core.Parser[X, String, Char]` and `A = Foo`, because then `F` would be an instances of all sorts of type classes. Therefore I suggest to change the order of the type parameters to `Parser[Elem, Repr, +T]`. It just makes sense given the direction in which the Scala language seems to be evolving.

mberndt123:
… to work better with -Ypartial-unification.

This fixes #176. I don't intend this to be merged in its current form as there are a bunch of other type constructors whose parameter order should also be changed, i. e. all subtypes of Parser, `Parsed` and its subtypes and more. But before I make those changes I'd like to find out if there is any support for the idea, hence the PR.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -123,7 +123,9 @@ lazy val fastparse = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .jvmSettings(fork in (Test, run) := true)
   .nativeSettings(nativeSettings)
 lazy val fastparseJS = fastparse.js
-lazy val fastparseJVM = fastparse.jvm
+lazy val fastparseJVM = fastparse.jvm.settings(
+  mimaPreviousArtifacts := Set("com.lihaoyi" %% "fastparse" % "1.0.0")
+)
 lazy val fastparseNative = fastparse.native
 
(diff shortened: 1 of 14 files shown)


lihaoyi:
Out of date due to 2.0.0

PR #185, opened 2018-05-18
Add Automatic-Module-Name for Java 9 compatibility

### LINKED ISSUE #184: Add a 'Automatic-Module-Name' for Java 9 compatibility
Please provide an [`Automatic-Module-Name`](http://branchandbound.net/blog/java/2017/12/automatic-module-name/) in the JAR manifests.

If no module name is explicitly defined, Java 9 uses the JAR file name as the module name. With the Scala naming scheme, the module name ends with ".2.12", which is invalid.
This makes this library incompatible with Java 9 at runtime.

florian-beetz:
This adds an [`Automatic-Module-Name`](http://branchandbound.net/blog/java/2017/12/automatic-module-name/) to the JAR manifests, allowing the library to be used on the Java 9 module path.

If the library was used before on the Java 9 module path, the module name was derived from the filename of the JAR at runtime to `sourcecode.2.12`, `sourcecode.utils.2.12`, etc., which are not a valid Java identifiers.

Fastparse is compatible with Java 9 when the dependency on sourcecode was updated to a version where https://github.com/lihaoyi/sourcecode/pull/49 is included.

Closes #184

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -50,7 +50,11 @@ val shared = Seq(
           <name>Li Haoyi</name>
           <url>https://github.com/lihaoyi</url>
         </developer>
-      </developers>
+      </developers>,
+  // Java 9 settings
+  packageOptions in (Compile, packageBin) += Package.ManifestAttributes(
+    "Automatic-Module-Name" -> s"com.lihaoyi.${name.value.replace('-', '.')}"
+  )
 )
 
 lazy val nativeSettings = Seq(


florian-beetz:
@lihaoyi Do you think we could get this (and https://github.com/lihaoyi/sourcecode/pull/49) merged? This is a blocker for Java 9 compatibility for us right now.

aborg0:
This test failure seems unrelated. I see similar test failures locally too (Windows 10 64 bit, OracleJDK 1.8.0_120 and also with 1.8.0_144 run from sbt with the `fastparseJVM/test`)

tobiasdiez:
@lihaoyi @olafurpg @vovapolu can you please merge this PR (or fix the underlying issue in another way), because otherwise the library cannot be used with Java 9 or newer. Thanks a lot!

tobiasdiez:
I tried to update this PR to reflect the recent changes to the build script. However, I was not able to find an option in mill to add something to the MANIFEST.MF file. 

It appears that https://github.com/lihaoyi/mill/blob/eececd6bbabd88b155a86e376a19d3aeee19cccf/main/src/modules/Jvm.scala#L193-L201 is responsible for writing the meta file, but the content does not seem to be configurable. @alexarchambault @olafurpg any ideas how to proceed to get a version that works with >= Java 9?

tobiasdiez:
@lihaoyi Is this PR still relevant for 2.2.3?

PR #186, opened 2018-05-22
Minor changes needed for dotty compatibility

OlivierBlanvillain:
I've just added fastparse to the dotty community build, here is the full diff in case you are curious: https://github.com/dotty-staging/fastparse/commits/dotty (I've destroyed the build and commented out all the macros for now...)

This PR cherry picks 3 commits that I believe could be merged as is as minor improvements.


allanrenucci:
@lihaoyi FWIW The test failure is unrelated to this PR. I can reproduce on master with:
```
> scalaparseJVM/testOnly -- scalaparse.ProjectTests.scala/scala
```
This is due to a recently added [test](https://github.com/scala/scala/blob/2.13.x/test/files/run/productElementName.scala) in `scala/scala`

lihaoyi:
Out of date due to 2.0.0

PR #189, opened 2018-07-15
Seq sequence support

aborg0:
Adds support for `sequence`. It allows multiple parsers (with same result type) to be combined with possibly a(n ignored) separator parser, just like in `rep`.

Feel free to suggest changes.


lihaoyi:
looks good to me

PR #204, opened 2018-11-28
Check that parameters used in parsers are by-name (or of wildcard type)

LPTK:
The right way of doing parameterized parsers is to use by-name parameters, as in:

```scala
def twice[T, _: P](p: => P[T]): P[(T,T)] = p ~ p
```

However, currently nothing prevents users from not using a by-name parameter. And nothing in the API seems to imply that parser parameters should be by-name.

I made the mistake myself, and was bitten by a class-cast exception coming out of nowhere. If you're lucky, you'll get a strange warning saying "a pure expression does nothing in statement position" which hints at a problem – but you won't necessarily get that warning, and it's not sufficient for users to know how to solve the problem anyway.

This PR augments the fastparse macros with a checking pass that makes sure the parsers that come from parameters are from by-name parameters.

It's not a complete check by any means and can be subverted, but I think it's much better than nothing.

### DIFF
--- a/fastparse/src/fastparse/internal/MacroImpls.scala
+++ b/fastparse/src/fastparse/internal/MacroImpls.scala
@@ -16,23 +16,32 @@ import scala.reflect.macros.blackbox.Context
   */
 object MacroImpls {
   
-  def checkByNames[T](c: Context)(p: c.Expr[T]): Unit = {
+  /** Checks a tree for subexpressions that contain references to parsers passed as parameters,
+    * and makes sure these parameters are by-name.
+    * We exclude from the check:
+    *  - implicit parameters, because they are used pervasively to pass context around, as in: `def foo[_: P] = ...`
+    *  - parsers whose type argument is an existential/wildcard, because they can be used safely (their value is not used)
+    */
+  private[fastparse] def checkByNames[T](c: Context)(p: c.Expr[T]): Unit = {
     import c.universe._
     val ParsingRunSym = typeOf[ParsingRun[_]].typeSymbol
-    
+
     // Note: it is not enough to just check the top-level tree, as sometimes it may be wrapped in EagerOps or others.
-    
-    new Traverser {
+
+    new Traverser{
       override def traverse(tree: Tree): Unit = {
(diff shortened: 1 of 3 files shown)


LPTK:
Hey @lihaoyi. Thoughts about moving forward with this?

lihaoyi:
@LPTK haven't had time to properly review it yet; do you mind if I just leave this here for a while?

LPTK:
Not at all, no worries 😉

andyczerwonka:
> @LPTK haven't had time to properly review it yet; do you mind if I just leave this here for a while?

This sounds like the same issue I've reported on the Gitter channel.
image

milanvdm:
@lihaoyi Any progress on this?
Ive got the exact same issue that @andyczerwonka is describing: `StringIn(listOfStrings:_*)` fails with `Function can only accept constant singleton type`

crakjie:
As all of you tried `StringInIgnoreCase` which does not works.

The `listOfString.map(s=>IgnoreCase(s)).reduce(_ | _)` which does not work this time because of a scalac that have difficulties to works with macro in some conditions.
(I'm not an expert but it's probably related to https://github.com/scala/bug/issues/12112)

So as work around I used the `filter` solution which is ugly and not efficient.
```scala
  private def operatorInSet[_: P](set : Set[String] ): P[String] = {
    val maxRep = set.map(_.size).max - 1
    P(
      ((letter) ~ (letter  | space).rep(max = maxRep)).!.filter(v=>set.contains(v.toUpperCase))
    )
  }
  ```
  
  It's works only because I store my set as upper cases.

He-Pin:
@LPTK Any update?

LPTK:
@He-Pin I don't know what's the current status of this problem. I no longer use fastparse myself. The code in this PR was written on 2018 and never reviewed by @lihaoyi.

He-Pin:
@LPTK I think that's because there well need a Scala 3 part too now ? thanks for the update.

LPTK:
Would be nice if there was a check there as well I guess (haven't tried the Scala 3 version), but it shouldn't be a requirement for adding this check to the Scala 2 check.

PR #220, opened 2019-04-26
Switch to scala 2.13.0-RC1

alexarchambault:


alexarchambault:
https://github.com/lihaoyi/mill/pull/593 may fix that

PR #221, opened 2019-04-26
Release from CI

alexarchambault:


.travis.yml
```diff
@@ -1,5 +1,9 @@
 sudo: false
 language: scala
+stages:
+  - name: test
+  - name: release
+    if: (branch = master AND type = push) OR (tag IS present)
 matrix:
```
olafurpg: I think this needs to be `stages` to use stages

.travis.yml
```diff
@@ -26,6 +30,12 @@ matrix:
   - env: CMD="mill demo.fullOpt && sbt readme/run"
     jdk: oraclejdk8
 
-script:
+  - stage: release
+    script:
+      - echo "$PGP_SECRET" | base64 --decode | gpg --import
+      - mill __.publish --sonatypeCreds "$SONATYPE_USERNAME:$SONATYPE_PASSWORD" --release true --gpgPassphrase "$PGP_PASSPHRASE"
```
olafurpg: Will this run on every merge into master or only for tag pushes?

alexarchambault: Every merge. But things are automatically pushed to the right repo (releases or snapshots).

.travis.yml
alexarchambault: You mean `jobs`? I was under the impression that `jobs` and `matrix` were equivalent (and [this](https://travis-ci.org/lihaoyi/fastparse/builds/524877277) goes in that sense).

alexarchambault: Just renamed it to `jobs` just in case.

.travis.yml
```diff
@@ -1,6 +1,10 @@
 sudo: false
 language: scala
-matrix:
+stages:
+  - name: test
+  - name: release
+    if: (branch = master AND type = push) OR (tag IS present)
+jobs:
```
olafurpg: :+1:

.travis.yml
olafurpg: I meant `jobs` yes, but maybe it's equivalent to `matrix`, I don't know 🤔

olafurpg:
Let's see if the snapshot release works 🍿 https://travis-ci.org/lihaoyi/fastparse/jobs/524888327

olafurpg:
Looks like a success 🎉 

```
coursier fetch com.lihaoyi:fastparse_2.13.0-RC1:2.1.0-8+aae1dca-SNAPSHOT -r sonatype:snapshots
```

alexarchambault:
✨ on first try ✨

olafurpg:
The release for 2.1.1 timed out on travis 😭  https://travis-ci.org/lihaoyi/fastparse/jobs/524928811

Should we try to split the release stage into parallel jobs for 2.11,2.12,2.13?

PR #251, opened 2021-04-23
Implement P.collect

runtologist:
Transforms the result of this parser using the given partial function, failing the parse if the partial function is not defined on the result of the current parser. This is eqivalent to `.filter(f.isDefinedAt).map(f.apply)`.


lihaoyi:
dummy commit to trigger CI

PR #253, opened 2021-06-14
Make Failure extractor exhaustive

LPTK:
This makes code like the following compile without an exhaustiveness warning:

```scala
parse(...) match {
  case Failure(err, index, extra) =>
    ...
  case Success(p, index) =>
    ...
}
```

Instead, one currently has to write:

```scala
parse(...) match {
  case f: Failure =>
    val Failure(err, index, extra) = f
    ...
  case Success(p, index) =>
    ...
}
```


LPTK:
@lihaoyi care to merge this? It's very minor.

lihaoyi:
pushed a dummy commit to retrigger CI

LPTK:
Thanks!

PR #259, opened 2021-11-13
Add mill-mima to check binary backward compatibility

lolgab:


lolgab:
Binary compatibility was broken in #253 but I'm merging this either way.

PR #264, opened 2022-12-16
Update Scala 2.13, Mill, Scala.js, Native and Actions

lolgab:


lolgab:
Merging since the failure in test-jvm was already there and the PR did not make it worse.

PR #267, opened 2022-12-27
Use `[_p: P]` instead of `[_: P]` for Scala 3 compatibility

lolgab:
Let's see if we can shrink #262 so it's easier to review by merging first the easy changes and rebasing, so the diff is smaller.


davidrwood:
Is this essentially competing with https://github.com/com-lihaoyi/fastparse/pull/266 ?

davidrwood:
Ah! I see now that it's not. I hadn't noticed you'd merged it already.

lolgab:
I wanted to help the migration by merging the easy syntactic changes first so the main PR becomes smaller and smaller.

PR #271, opened 2023-02-28
Scala 3 support

lihaoyi:
All tests pass on Scala 3, on all platforms (JVM/JS/Native), and on all old versions of Scala.

Pulls in a bunch of work by @lolgab (https://github.com/com-lihaoyi/fastparse/pull/252) @rmgk (https://github.com/com-lihaoyi/fastparse/pull/262) and @ajrnz (https://github.com/com-lihaoyi/fastparse/pull/266)

# Notable Changes:

1. Moved MIMA checking only to the core `fastparse` modules; I don't think we can promise the same stability guarantees for the example parsers 
2. Tweaked the source file config in each module to allow per-platform-per-version sources, and extended that ability to test sources.
3. All the `.map(Foo)` calls have to become `.map(Foo.apply)`, `.map(Foo.tupled)` calls have to become `(Foo.apply _).tupled`
4. Duplicated `package.scala` for Scala 2 and 3, but splitting out the shared non-macro logic into `SharedPackageDefs.scala`
5. All macros had to be re-implemented in `fastparse/src-3/`
6. All implicits had to be given explicit return types
7. Fixed a bug in `aggregateMsgInRep` where we were not properly propagating failure strings
8. `then` has to be back-ticked
9. Replaced all `scala.Symbol`s in PythonParse with `String`s
10. Replaced the `_` method in ScalaParse with `Underscore`
11. Replaced some residual usage of uTests' `'foo -` syntax with `test("foo") -`

# Performance 
Performance seems to have taken about a 10% hit in Scala 3. Probably missing out on some optimizations that we do in Scala 2. I'm not super familiar with how scala 3 macros work, clawing back the 10% can come in a follow up PR

Scala 2 Bench
```
------------------ Running Tests perftests.string.ScalaParse ------------------
ScalaParse Benchmark
Max time - 10000 ms. Iterations - 5.
Iteration 1
Benchmark 0. Result: 984
Benchmark 1. Result: 82
Iteration 2
Benchmark 0. Result: 632
Benchmark 1. Result: 88
Iteration 3
Benchmark 0. Result: 618
Benchmark 1. Result: 96
Iteration 4
Benchmark 0. Result: 625
Benchmark 1. Result: 99
Iteration 5
Benchmark 0. Result: 596
Benchmark 1. Result: 99
984 82
632 88
618 96
625 99
596 99
```

Scala 3 Bench
```
------------------ Running Tests perftests.string.ScalaParse ------------------
ScalaParse Benchmark
Max time - 10000 ms. Iterations - 5.
Iteration 1
Benchmark 0. Result: 899
Benchmark 1. Result: 62
Iteration 2
Benchmark 0. Result: 535
Benchmark 1. Result: 86
Iteration 3
Benchmark 0. Result: 545
Benchmark 1. Result: 85
Iteration 4
Benchmark 0. Result: 547
Benchmark 1. Result: 86
Iteration 5
Benchmark 0. Result: 538
Benchmark 1. Result: 85
899 62
535 86
545 85
547 86
538 85
```


lihaoyi:
@lolgab looks like it's been exactly a year to the day you started this: https://github.com/com-lihaoyi/fastparse/pull/271/commits/624c8131fabd8112180675281b21b89a84d5df01 landed 1st March 2022, and here we are 1st March 2023. Now just need to do final review and cleanups before we merge/release this

reid-spencer:
Thank you all for managing to get this to the finish line.

lolgab:
Thank you for taking over @lihaoyi :)
There are some small details that can be fixed in the future, so I'm more than happy to release a new fastparse 2.x version with the long awaited Scala 3 support.
Good call on restricting Mima to the fastparse module only.
The little details I'm thinking about are the usage of package object in Scala 3 (which I think is deprecated) and the `.rep` which become `.rep()`, if you read the original PR code and description, there were some overloads that didn't work on Scala 3, but the linked issue is now closed, I'm wondering if they can be reintroduced now so in Scala 3 you can also do `.rep`.
Anyways, I don't think these should block the release, since we waited already for too long.

lihaoyi:
Looks like bare `.rep` and `.repX` work now. Not sure when that happened, but I reverted the extra `()`s we added and things still seem to pass

lolgab (approved):
LGTM!
Thank you @lihaoyi for taking this to the finish line! And, of course, thank you to @rmgk for all the work to port the macro code to Scala 3!
Congratulations 🎉
Looking forward to have complete Scala 3 support in Ammonite and Mill after this!

fastparse/src-3/fastparse/internal/MacroRepImpls.scala
```diff
@@ -0,0 +1,109 @@
...
+  )(repeater: Expr[Implicits.Repeater[T, V]], ctx: Expr[ParsingRun[_]])(using quotes: Quotes): Expr[ParsingRun[V]] = {
+    import quotes.reflect.*
+
+    '{
+
+      val ctx1        = $ctx
+      val repeater1   = $repeater
+      var originalCut = ctx1.cut
+      val acc         = repeater1.initial
+      @ _root_.scala.annotation.tailrec
```
smarter, line 32: FWIW, you don't need to bother with fully-qualified names in Scala 3 macros, they're hygienic.

reid-spencer: LGTM!

build.sc
```diff
@@ -5,21 +5,25 @@ import scalanativelib._
...
-import $ivy.`de.tototec::de.tobiasroeser.mill.vcs.version::0.1.4`
+
+import mill.scalalib.api.Util.isScala3
+import $ivy.`de.tototec::de.tobiasroeser.mill.vcs.version::0.3.0`
+
 import de.tobiasroeser.mill.vcs.version.VcsVersion
 import $ivy.`com.github.lolgab::mill-mima::0.0.13`
 import com.github.lolgab.mill.mima._
 
+val scala31 = "3.1.3"
```
reid-spencer, line 16: Can we get to 3.2.2?

reid-spencer (approved):
LGTM!

Thanks all who've been involved in making this happen. This is a big deal! Much appreciated!

lihaoyi:
@lolgab shall we tag a new version now that this is merged? We can perform further optimizations and improvements in follow ups

lolgab:
> @lolgab shall we tag a new version now that this is merged? We can perform further optimizations and improvements in follow ups

We can.
@lefou Since we bumped geny and other dependencies to a new major, this should also be a new major, right?
In this case I would check if your next PR is backward compatible, to avoid having to release a 4.0.0 right after. To do so locally you can publish the lib locally, change `mimaPreviousVersions` to the snapshot version, apply your next PR and  then run mimaReportBinaryIssues on the Scala 2 modules.

lihaoyi:
If it's a new major, we should try to squeeze in more breaking changes if possible. e.g. the whitespace PR I have open is definitey binary incompatible

PR #272, opened 2023-03-01
Make Whitespace a proper trait rather than function to avoid unwanted implicit conversions

lihaoyi:
Fixes https://github.com/com-lihaoyi/fastparse/issues/261, which is caused by the whitespace `P[_] => P[Unit]` implicits firing as implicit conversions rather than the implicit parameters they were originally intended to be. Since whitespace is meant to ignore failures, this caused failures caused by the `Fail.! : P[String]` to be ignored when implicitly converted to `P[Unit]`

The fix is to create a proper `Whitespace` trait to inherit from. 

Note that this is a binary incompatible change. It's stacked on top of https://github.com/com-lihaoyi/fastparse/pull/271 for convenience, but should probably land separately after. The relevant changes are in this commit https://github.com/com-lihaoyi/fastparse/pull/272/commits/d87ab6d9de062ec1e0bbf897643cefef7cd03338 if anyone wants to look.

Note that I moved the custom whitespace tests to a `scala2.12+` folder to skip them on 2.11. 2.11 does not support SAM conversion, which makes defining custom whitespaces a lot more boilerplatey. Can still be done, but no need to burden everyone with boilerplatey examples just to cater for the 2.11 folks in the test suite. Things are unlikely to break just for 2.11 anyway

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -205,7 +205,11 @@ trait CommonCrossModule extends CrossScalaModule with PublishModule {
   def platformSegment: String
   def millSourcePath = super.millSourcePath / os.up
   def sources = T.sources {
-    super.sources().flatMap{p => Seq(p, PathRef(p.path / os.up / s"${p.path.last}-$platformSegment")) }
+    super.sources()
+      .flatMap{p => Seq(p, PathRef(p.path / os.up / s"${p.path.last}-$platformSegment")) } ++
+      (if (scalaVersion() != scala211) {
+        Seq(PathRef(millSourcePath / "src-2.12+"))
+      } else Seq())
   }
 }
 trait CommonTestModule extends ScalaModule with TestModule.Utest{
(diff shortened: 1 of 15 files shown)


lihaoyi:
@lefou @lolgab if we expect the next version of Fastparse to be 3.0.0, then I'll merge this to get it in before the version bump

lefou:
Can you rebase this PR on top of #271, please?

lefou:
> @lefou should already be rebased

I meant on top of `master` branch.

lihaoyi:
@lefou I squashesd it on top of latest master, we can use the `Squash and Merge` button to make sure the PR description is used on the final commit that ends up in master

lefou (approved):
Looks good to me.

lefou (changes requested):
The change looks good to me. But there are test failures now.

lefou (approved):

PR #273, opened 2023-03-03
Scala 3 perf improvements, consolidate rep implementations

lihaoyi:
1. Previously `.rep()` never inlined, and `.repX()` would always inline and fail with a compile error if it was not possible. This should make both perform inlining at a best-effort level, giving us the best of both worlds

2. I saw a lot of calls to `List()` turning up in my JProfile, replaced them with raw `::` calls. Seems like this is optmized automatically in Scala 2, but not in Scala 3 https://github.com/lampepfl/dotty/issues/17035

3. Lift whitespace NoWhitespace check in `.rep` to compile time, like we do in `~`.

4. Combine all the menagerie of `.rep` implementations into one Scala 3 macro. Scala 2 is still a bit of a mess, cleaning that up is left for future

6. Inlined the character checking in `literalStrMacro` to avoid allocating an intermediate function to call

Seems to provide a small but measurable performance boost for successful parses, and a significant performance boost for failures (where we allocate a lot more lists as part of error reporting). Not quite enough to catch up to Scala 2 performance, but brings us maybe half way there.

Scala 3 before:

```
JsonnetParse Benchmark
Max time - 10000 ms. Iterations - 5.
Iteration 1
Benchmark 0. Result: 9331
Benchmark 1. Result: 1546
Iteration 2
Benchmark 0. Result: 7715
Benchmark 1. Result: 1947
Iteration 3
Benchmark 0. Result: 7549
Benchmark 1. Result: 1976
Iteration 4
Benchmark 0. Result: 7613
Benchmark 1. Result: 1953
Iteration 5
Benchmark 0. Result: 7686
Benchmark 1. Result: 1907
```

Scala 3 after:

```
JsonnetParse Benchmark
Max time - 10000 ms. Iterations - 5.
Iteration 1
Benchmark 0. Result: 9466
Benchmark 1. Result: 2611
Iteration 2
Benchmark 0. Result: 8152
Benchmark 1. Result: 2832
Iteration 3
Benchmark 0. Result: 8139
Benchmark 1. Result: 2799
Iteration 4
Benchmark 0. Result: 8020
Benchmark 1. Result: 2844
Iteration 5
Benchmark 0. Result: 8126
Benchmark 1. Result: 2868
```

Scala 2.13:

```
JsonnetParse Benchmark
Max time - 10000 ms. Iterations - 5.
Iteration 1
Benchmark 0. Result: 9850
Benchmark 1. Result: 2773
Iteration 2
Benchmark 0. Result: 8781
Benchmark 1. Result: 2912
Iteration 3
Benchmark 0. Result: 8742
Benchmark 1. Result: 2916
Iteration 4
Benchmark 0. Result: 8782
Benchmark 1. Result: 2912
Iteration 5
Benchmark 0. Result: 8703
Benchmark 1. Result: 2940
```


build.sc
```diff
@@ -13,7 +13,7 @@ import de.tobiasroeser.mill.vcs.version.VcsVersion
 import $ivy.`com.github.lolgab::mill-mima::0.0.13`
 import com.github.lolgab.mill.mima._
 
-val scala31 = "3.1.3"
```
lefou, line 16: Do you drop Scala 3.1 support on purpose?

lihaoyi: Yeah, 3.1 gives some weird `unable to call macro trait in same file` errors that go away on 3.22

lolgab (approved):

PR #274, opened 2023-03-03
Attempt to share macro .rep logic in Scala3

lihaoyi:
This PR attempts to consolidate all the macros and runtime implementations into one shared code path

Still has some performance hit v.s. the status quo, so not ready to merge


lihaoyi:
superseded by https://github.com/com-lihaoyi/fastparse/pull/273

PR #277, opened 2023-03-07
Add sourceMap mapping to Scala.js

lolgab:
This makes source maps to point to Github instead of a non-existing local source file (with the path from the CI machine)

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -38,6 +38,16 @@ object fastparse extends Module{
     def platformSegment = "js"
     def millSourcePath = super.millSourcePath / os.up
     def scalaJSVersion = crossScalaJsVersion
+    private def sourceMapOptions = T.task {
+      val vcsState = VcsVersion.vcsState()
+      vcsState.lastTag.collect {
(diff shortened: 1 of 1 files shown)


lolgab:
I tested it manually by tagging locally a commit, publishing locally, and then importing the library in a scala-cli script.
Looking at the generated source map it contains correct links:
```json
"sources": ["https://raw.githubusercontent.com/scala-js/scala-js/v1.12.0/javalib/src/main/scala/java/lang/Integer.scala", "https://raw.githubusercontent.com/lihaoyi/fastparse/3.0.0/fastparse/src/fastparse/Parsed.scala", "https://raw.githubusercontent.com/lihaoyi/fastparse/3.0.0/fastparse/src/fastparse/ParsingRun.scala"
```
Consider that this code was ported [from upickle](https://github.com/com-lihaoyi/upickle/pull/432/files) and it was also manually tested there.

lolgab:
Merged so we don't forget to add it to the upcoming release.

PR #281, opened 2023-03-24
Squelch useless warnings on `.rep`

lihaoyi:
Fixes https://github.com/com-lihaoyi/fastparse/issues/279

I added the appropriate `@nowarn` annotation in the macro. 

Tested by enabling `-Xfatal-warnings`. For now I only do this in Scala 2.13, as Scala 2.11/2.12 have no `@nowarn` (I had to leave a stub for them) and Scala 3.x (a) doesn't have the same warning and (b) somehow has a bunch of other `is more specialized than the right hand side` warnings that `@nowarn` doesn't seem to squelch

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -229,7 +229,7 @@ trait CommonTestModule extends ScalaModule with TestModule.Utest{
     ivy"com.lihaoyi::utest::0.8.1",
   )
   override def scalacOptions = super.scalacOptions() ++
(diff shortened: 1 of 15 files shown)


luigimarano:
Unfortunately I think this change has been causing problems with my build. I now get the following error message:

```@nowarn annotation does not suppress any warnings```
scala version 2.13.10

Given that it didn't happen with v`3.0.0` I'm just going to stick with that version for now

gipeshka:
Can confirm that update to `3.0.1` fails with the same message for me.

Issue #296, opened 2023-11-22
Fail ignores its argument and uses "fail"


reid-spencer:
The `Fail` operator appears to ignore its documentation and always produces a failure message of "fail". The documentation shows this example:
```scala
def failWithLabel[$: P] = P( Fail("custom fail msg") )
val Parsed.Failure(_, 0, extra) = parse("asdad", failWithLabel(_))
assert(extra.trace().longMsg == """Expected failWithLabel:1:1 / fail:1:1, found "asdad"""")
```
I believe the intent of using `Fail` within `failWithLabel`,  based on its name, is to provide a new failure label. However, the assertion on the last line shows the issue. I believe it should say:
```scala
assert(extra.trace().longMsg == """Expected failWithLabel:1:1 / custom fail message:1:1, found "asdad"""")
```
If you look at [this line of code](https://github.com/com-lihaoyi/fastparse/blob/master/fastparse/src/fastparse/SharedPackageDefs.scala#L164 ) you can see the mistake:
```scala 
/**
   * No-op parser with a custom error message that always fails, consuming zero characters
   */
  def Fail(msg: String)(implicit ctx: P[_]): P[Nothing] = {
    val res = ctx.freshFailure()
    if (ctx.verboseFailures) ctx.reportTerminalMsg(ctx.index, () => "fail")   <===== "fail" should be msg
    res
  }
```
Changing `"fail"` (static message) to `msg` will make use of the message and provide much more significant value in reporting custom error messages in fastparse.

reid-spencer:
@lihaoyi - Locally, running the tests succeed. In the CI, they fail because of the return code (1 instead of 0). The only difference I can see is hardware and the JVM is 17 not 8.  Do you have a "must pass all tests in CI" policy before merging?

lihaoyi:
Yes, we want CI to be green before merging. The last CI run on master was green, so I would expect any PRs to be green as well

reid-spencer:
> Yes, we want CI to be green before merging. 

Good, I concur. 

> The last CI run on master was green, so I would expect any PRs to be green as well

I've merged nothing to main, yet. See PR #297 and its various failures, if you're interested. 

I'll continue to look for solutions.

PR #297, opened 2023-11-22
Implement fix for #296

reid-spencer:
Simply replace '"fail"' with 'msg' to use the provided custom error message

Additionally:
* Add 3.3.1 as a cross-build target
* Put quotes in the "Common Commands" section of the README so that the commands are directly 
  copyable for Zsh, Csh, Bash, etc. Should work on Windows too

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,18 +46,18 @@ compiled to Scala.js. These all live in `demo/`.
 Common Commands
 ---------------
 
-- `mill -w fastparse.jvm[2.12.10].test` runs the main testsuite. If you're
+- `mill -w "fastparse.jvm[2.12.10].test"` runs the main testsuite. If you're
   hacking on FastParse, this is often where you want to go
 
 - You can run the other suites via `fastparse.js`, `scalaparse.jvm`, etc. if you
   wish, but I typically don't and leave that to CI unless I'm actively working
   on the sub-project
 
-- You can use `mill -w fastparse.jvm[_].test` to run it under different Scala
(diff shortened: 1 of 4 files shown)


lihaoyi:
Seems like there's some issues with integration tests missing files; not sure why that's happening, given the tests use pinned SHAs for the external repositories they clone.

reid-spencer:
> Seems like there's some issues with integration tests missing files; not sure why that's happening, given the tests use pinned SHAs for the external repositories they clone.

Yes. I have been able to find the errors now (lots of output!) and confirm two errors in th test repositories for scalaparse:
`apache/kafka`. I checked the SHA in the test case and it exists in github repository so apache hasn't rebuilt the repo or anything.  

🤷🏼‍♂️

lihaoyi:
Sounds good. Regardless of why, it's probably unrelated to this PR

reid-spencer:
I concur. My sequence of events for this closed PR are:
* the full test suite ran with the last commit, there were several failures
* re-ran those failing ones, all passed but `test-jvm`
* re-ran `test-jvm` and it passed
* the net result is visible above in the `View Details` button (all tests passed)
* I merged the commit
* The `master` branch build failed on `test-jvm`
* I re-ran `test-jvm` and it failed again

So, it seems like github is having some inconsistency with accessing the test source repositories. 

In any event, the `master` branch now has the fix on it but it won't auto-release as a result of the inconsistency. I'm okay with using a local build. I'll file an issue for delving into this CI/build issue more significantly.

PR #310, opened 2024-05-18
Nowarn fix for `Whitespace` macro

He-Pin:
Motivation:
refs: https://github.com/com-lihaoyi/fastparse/issues/285
with reading: https://www.scala-lang.org/2021/01/12/configuring-and-suppressing-warnings.html

Modification:
1. I updated some depenecies,both scala and etc
2. because of https://github.com/com-lihaoyi/mill/issues/3167, I have to comments out the native modules locally
3. add `-Xlint:unused` to reproduce the problem
4. updated some tests when `3` is enabled.
5. remove 2.11 related code
6. as Scala 2.12.x now supports the nowarn , so I changed the code to use the `@nowarn` annotation directly.
7. the root cause is the annotation in `WhiteSpace` macro

Result:
I think the issue is fixed now, and as I'm using Fastparse within a Java project, so I was not knowing this issue.
All tests passed locally.


He-Pin:
@fanf Would you like to give this a try? I mean locally.

lihaoyi:
The PR looks good to me, once you get CI green I can merge it and close out the bounty

He-Pin:
@lihaoyi Would you like give it another round of run, thanks.

He-Pin:
@lihaoyi The CI is happy now:)

Note: in Scala 2.12, I added two mina filters where the nowarn is the scala nowarn now.

lihaoyi:
@He-Pin can we leave the original `nowarn` in place but deprecated? That way we can avoid adding the mima filters

He-Pin:
@lihaoyi OK, I think that would be better.

He-Pin:
@lihaoyi I have updated and removed the  added mima.

scalaparse/test/src-2.13-jvm/scalaparse/ScalacParser.scala
```diff
@@ -0,0 +1,61 @@
+package scalaparse
```
lihaoyi, line 1: @He-Pin is this change intentional?

He-Pin: Yes, because the api in nsc is different in Scala 2.12 and newer version of scala 2.13.

scalaparse/test/src-2.13-jvm/scalaparse/ScalacParser.scala
```diff
@@ -0,0 +1,61 @@
...
+          fail = true
+        }
+        override def incompleteInputError(off: Offset, msg: String) = {
+          fail = true
+        }
+      }
+
+      override def incompleteInputError(msg: String, actions: List[CodeAction]): Unit = {
+        fail = true
+        super.incompleteInputError(msg, actions)
```
He-Pin, line 50: the new api has a `actions`

scalaparse/test/src-2.13-jvm/scalaparse/ScalacParser.scala
```diff
@@ -0,0 +1,61 @@
...
+      }
+
+      override def syntaxError(offset: global.syntaxAnalyzer.Offset, msg: String, actions: List[CodeAction]): Unit = {
+        fail = true
+        super.syntaxError(offset, msg, actions)
```
He-Pin, line 55: so does `syntaxError` method. otherwise, it will report `override nothing` error.

scalaparse/test/src-2.13-jvm/scalaparse/ScalacParser.scala
He-Pin: So I have to separate this test for Scala 2.12 and Scala 2.13, maybe I can split it into a dedicated PR. but as the commits is one by one, maybe that's fine too.

He-Pin: related link: https://github.com/scala/scala/pull/10406

lihaoyi:
@He-Pin not sure if I'm missing something, but where is the new version of ScalacParser used in the PR?

He-Pin:
@lihaoyi Because I updated the Scala 2.13.x to 2.13.14 in this PR, and with https://github.com/scala/scala/pull/10406 , the additional `actions: List[CodeAction]` parameter is added, so when do cross compiling with Scala 2.12 and Scala 2.13, it will raise an error `override nothing` without this change in the scalaparse tests.

build.sc
```diff
@@ -13,11 +13,11 @@ import $ivy.`com.github.lolgab::mill-mima::0.0.23`
 import de.tobiasroeser.mill.vcs.version.VcsVersion
 import com.github.lolgab.mill.mima._
 
-val scala3 = "3.3.1"
-val scala213 = "2.13.10"
-val scala212 = "2.12.17"
-val scalaJS1 = "1.12.0"
-val scalaNative04 = "0.5.0"
+val scala3 = "3.3.3"
+val scala213 = "2.13.14"
```
He-Pin, line 17: it's now 2.13.14, and the nsc is updated https://github.com/scala/scala/pull/10406

lihaoyi:
@He-Pin got it, looks good! Can send me details at haoyi.sg@gmail.com and I'll close out the bounty

He-Pin:
thanks, I just implement a jsonpath (https://www.rfc-editor.org/rfc/rfc9535#name-collected-abnf-grammars) with Fastparse at work, it works great.

PR #312, opened 2024-05-21
chore: Enable lint and fix some unused warning.

He-Pin:
Motivation:
Enable unused lint for fastparse

Modification:
Some straightforward changing to make compiler happy

Result:
less warning when comping.


fastparse/src-2/fastparse/package.scala
```diff
@@ -284,7 +284,6 @@ package object fastparse extends fastparse.SharedPackageDefs {
       val startTerminals = ctx.terminalMsgs
       parse0()
       ctx.noDropBuffer = oldNoCut
-      val msg = ctx.shortMsg
```
He-Pin, line 287: These two is removed.

PR #323, opened 2024-12-08
chore: Add withFilter

He-Pin:
Motivation:

Make if guard works


fastparse/src-2/fastparse/package.scala
```diff
@@ -102,6 +102,12 @@ package object fastparse extends fastparse.SharedPackageDefs {
     def filter(f: T => Boolean)
               (implicit ctx: P[Any]): P[T] = macro MacroImpls.filterMacro[T]
 
+    /**
+     * Used by for-comprehensions.
+     */
+    def withFilter(f: T => Boolean)
+                  (implicit ctx: P[Any]): P[T] = macro MacroImpls.filterMacro[T]
+
```
russwyte, line 110: Why not just make this call filter?  Otherwise one might need to impl twice.

He-Pin: We can't, it compiles an error.

PR #324, opened 2024-12-08
chore: bump scala versions

He-Pin:
Makes it compiles on Java 21


He-Pin:
```
rror: Exception in thread "main" java.lang.NoSuchMethodError: java.util.StringTokenizer.asIterator()Ljava/util/Iterator;
	at mill.eval.Plan$.splitEnclosing$1(Plan.scala:52)
	at mill.eval.Plan$.$anonfun$assignOverridenTaskSegments$3(Plan.scala:58)
	at scala.collection.immutable.List.map(List.scala:247)
	at scala.collection.immutable.List.map(List.scala:79)
```
@lihaoyi Seems mill 0.12.4 need a fix then, because it now using the method only exists from Java 9.

